PR #3, opened 2025-01-24
add HUD features. 

zjwhale:
Now, sim_scene.host_star_image is an HDU. A data folder has been created, and an example FITS file has been added to it. The code reads this FITS file and assigns its header to sim_scene.host_star_image. The default header can be customized by passing the header_info keyword to the create_hdu() function. If no value is provided (header_info=None), the header defaults to the one read from the example FITS file.

### DIFF
--- a/corgisim/instrument.py
+++ b/corgisim/instrument.py
@@ -1,5 +1,8 @@
 
-
+import proper
+import numpy as np
+from astropy.io import fits
+import roman_preflight_proper
 
 class CorgiOptics():
     '''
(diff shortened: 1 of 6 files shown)


corgisim/instrument.py
```diff
@@ -88,7 +89,7 @@ def get_psf(self, input_scene, on_the_fly=False, oversample = 1, return_oversamp
 
         sim_scene = scene.SimulatedScene(input_scene)
         
-        sim_scene.host_star_image =  image
+        sim_scene.host_star_image = create_hdu(image)
```
maxwellmb, line 92: So at this point I don't think we need the HDUs to have all this header information yet. Right now they're just data containers that whole the simulated scene. They're not yet full L1 data products; it's only when we start actually constructing the final simulation product that we want this full structure. For now I think we can just have something like `sim_scene.host_star_image = fits.PrimaryHDU(image)`, and then some header keywords that hold the input simulation settings.

maxwellmb (changes requested):

maxwellmb:
The comments here should be addressed in PR #5

PR #5, opened 2025-01-24
change the header_info in HDU features

zjwhale:
I've simplified the HDU features. The updated code now passes simulation-related keywords as headers to the HDU, including stellar properties and self.proper_keywords. However, I encountered a warning indicating that some keyword names are too long for a standard FITS header if copied directly from self.proper_keywords. While this works for now, we should discuss and finalize the header format at a later stage.

### One update: I used comment keyword to avoid the warning of long name for standard FITS header.

### DIFF
--- a/corgisim/test/test_on_axis_star.py
+++ b/corgisim/test/test_on_axis_star.py
@@ -56,7 +56,7 @@ def run_sim():
     image = sim_scene.host_star_image.data
 
     print('Final_intensity_get:', np.sum(image, dtype = np.float64))
-    #print(sim_scene.host_star_image[1].header)
+    #print(sim_scene.host_star_image.header)
 
     fig = plt.figure()
     plt.imshow(image)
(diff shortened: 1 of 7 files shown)


maxwellmb:
I think we'll want to merge in PR #4 first before addressing this one since many of the changes are overlapping.

corgisim/data/CGI_0000000000000000014_20221004T2359351_L1_.fits
maxwellmb, line 1: I need to double check that it's ok to have this file public.

corgisim/test/test_on_axis_star.py
```diff
@@ -0,0 +1,77 @@
...
+
+    image = sim_scene.host_star_image.data
+
+    print('Final_intensity_get:', np.sum(image, dtype = np.float64))
+    #print(sim_scene.host_star_image.header)
+
+    fig = plt.figure()
+    plt.imshow(image)
+    plt.colorbar()
+    plt.show()
```
maxwellmb, line 64: maybe let's add a kwarg to the function that toggles plotting. We won't want pytest to be plotting things, but we might want it when we're debugging.

setup.py
maxwellmb, line 1: Have you tested that this works?

maxwellmb (changes requested):
A few more comments on this one!

maxwellmb:
This PR is stale. Closing it without merging.

PR #12, opened 2025-03-10
add_emccd_features

zjwhale:
This PR request is about adding emccd simulation


corgisim/instrument.py
```diff
@@ -304,43 +305,183 @@ def inject_point_sources(self, scene, on_the_fly=False, oversample = 1, return_o
     
 class CorgiDetector(): 
     
-    def __init__(self, emccd_keywords):
+    def __init__(self, exptime ,emccd_keywords):
```
maxwellmb, line 308: lets move exptime to generate_detector_image since it's not a fundamental property of the detector (which is what the class represents.

corgisim/instrument.py
```diff
@@ -304,43 +305,183 @@ def inject_point_sources(self, scene, on_the_fly=False, oversample = 1, return_o
...
+        if 'cr_rate' in emccd_keywords: cr_rate = emccd_keywords['cr_rate']
+        if 'em_gain' in emccd_keywords: em_gain = emccd_keywords['em_gain']
+        if 'bias' in emccd_keywords: bias = emccd_keywords['bias']
+        if 'e_per_dn' in emccd_keywords: e_per_dn = emccd_keywords['e_per_dn']
+        if 'nbits' in emccd_keywords: nbits = emccd_keywords['nbits']
+        if 'numel_gain_register' in emccd_keywords: numel_gain_register = emccd_keywords['numel_gain_register']
+        if 'use_traps' in emccd_keywords: emccd_keywords = emccd_keywords['use_traps']
+        if 'date' in emccd_keywords: date = emccd_keywords['date']
+        if 'qe' in emccd_keywords:
+            raise Warning("Quantum efficiency has been added in the bandpass throughput; it must be enforced as 1 here.")
```
maxwellmb, line 463: You can do all of this as dictionaries: 
- put all the default parameters in a dictionary
- selectively replace the default, with anything new that's provided (see [here](https://github.com/roman-corgi/corgidrp/blob/08d49cf9252ff5445dc77278792d3eefe7657a5d/corgidrp/star_center.py#L141) for example
- pass in the dictionary to the function (keyword arguments can be passed into Python functions as dictionaries)

test/test_on_axis_star_on_dector_fullframe.py
maxwellmb, line 1: Tests should have assert statements that can be tested quantitatively. Although we don't have the CI set up yet with pytest, we can start preparing for it by having the tests in the correct form.

maxwellmb: This applies to both testing files

maxwellmb: One way to move forward is to let this through, but to make an Issue that these tests need to be modified. At least that way we're letting them through.

roman_preflight.py
maxwellmb, line 1: We probably don't want to track this on the github, right? Perhaps, we could add it to the git ignore?

corgisim/scene.py
```diff
@@ -272,6 +272,7 @@ def __init__(self, input_scene):
 
         #The following three attributes will hold astropy HDUs with the simulated images.
         self.host_star_image = None
+        self.host_star_image_on_detector = None
```
maxwellmb, line 275: we discussed removing this, right?

maxwellmb (changes requested):
Just a few changes here!

corgisim/instrument.py
zjwhale: Ok, did this

roman_preflight.py
zjwhale: Yes, I have added it to git ignore

corgisim/instrument.py
maxwellmb: Example from today (may not actually work), but gives you the idea. 

```
default_parameter_dictionary = {
    "full_well_serial": 100000.0,
    "full_well_image": 60000.0,
    "dark_rate": 0.00056,
    "cic_noise": 0.01,
    "read_noise": 100.0,
    "cr_rate": 0,
    "em_gain": 1000.0,
    "bias": 0,
    "pixel_pitch": 13e-6,
    "apply_smear": True,
    "e_per_dn": 1.0,
    "nbits": 14,
    "numel_gain_register": 604,
    "use_traps": False,
    "date": 2028.0
}

if input_dictionary is not None:
    for key in input_dictionary.keys():
        if key in default_parameter_dictionary:
            default_parameter_dictionary[key] = input_dictionary[key]
```

corgisim/scene.py
maxwellmb: Let's get rid of this and add a single self.detector_image property

corgisim/instrument.py
zjwhale: Just changed it

corgisim/scene.py
zjwhale: Yes, I removed `self.host_star_image_on_detector` and replaced it with `self.image_on_detector.`
Now, in the `CorgiDetector.generate_detector_image()` function, it checks whether `simulated_scene.host_star_image`, `simulated_scene.point_source_image`, and `simulated_scene.twoD_image` are None, and combines any that are not to generate the final detector image.

test/test_on_axis_star_on_dector_fullframe.py
zjwhale: add pytest to subframe simulations. But because cgisim does't generate full-frame image, we don't have pytest test in full frame simulation now

PR #23, opened 2025-04-07
Add installation instructions to readme

helloWorlds0:
Add instruction to readme. 
Also add a test_installation.py that copies roman_preflight.py and roman_preflight_compact.py into tests folders


maxwellmb (approved):
Let's wait to approve the CI PR first, so that we can have this one to be the first PR to go through CI testing.

PR #24, opened 2025-04-07
Setup continuous integration

helloWorlds0:


helloWorlds0:
I can't test the CI on any branch other than main

maxwellmb:
> I can't test the CI on any branch other than main

What should I do with this information?

helloWorlds0:
> > I can't test the CI on any branch other than main
> 
> What should I do with this information? 

Well, I can't really test if the CI works before merging it into main, so I am unsure if the current setup works. There are imo two possible ways of doing it: either I develop it on a separate branch, and merge it into main to see if it works, or I develop it directly into main (not really best practice but would allow me to debug much more quickly)

maxwellmb:
Gotcha. I'm not opposed to you working on main just to set this up, but now that I've properly set up the branch protections though, I think I'd have to set up something temporarily. I propose that I let it through right now and then if need be I can do that.

maxwellmb (approved):

PR #29, opened 2025-04-10
Generate and setup the documentation page

helloWorlds0:


helloWorlds0:
Setup for documentation page in readthedocs.io . However, in order to have this documentation, the owner of the repo has to be the one who sets it up.

helloWorlds0:
Modified so that it works with readthedocs.io 
Documentation available at [https://corgisim.readthedocs.io/en/latest/](https://corgisim.readthedocs.io/en/latest/)
One problem remaining in convolution.py : because of "import roman_phasec_proper" fails, the documentation doesn't go through it. Since I am unsure why we need this import in convolution.py and it's still in development, I am not fixing it right now.

maxwellmb:
> Modified so that it works with readthedocs.io Documentation available at https://corgisim.readthedocs.io/en/latest/ One problem remaining in convolution.py : because of "import roman_phasec_proper" fails, the documentation doesn't go through it. Since I am unsure why we need this import in convolution.py and it's still in development, I am not fixing it right now.

convolution.py was developed in the summer by an undergraduate, mostly in isolation of the eventual plan for the simulator. We're free to change what's needed and can comment out things if its causing problems. It's most related to what Alexis is working on, so you could consider coordinating with her. It also sounds not urgent, so we can maybe just fix it when the 2D convolution comes through.

maxwellmb (changes requested):
Is there a way to check to see if the documentation pages build without fail?

helloWorlds0:
https://app.readthedocs.org/projects/corgisim/builds/?utm_source=corgisim&utm_content=flyout but I don't know if you can see it or if it's only for my account
image

maxwellmb:
I can see it, but is there a way for us to check that they build correctly as part of the CI tests?

helloWorlds0:
I think so, based on [this](https://docs.readthedocs.com/platform/stable/pull-requests.html). Right now it builds on pull request on generate-and-setup-the-documentation-page but I'll change it to main once it's merged in.

maxwellmb (approved):

PR #32, opened 2025-04-10
add_tutorial_in_example_folder

zjwhale:
I add a tutorial jupyte notebook in example folder


maxwellmb:
Overall I think this is a good strategy for now as we bring people up to speed. In the not-too-distant future, we might split out some of the things like the install instructions to be their own pages on the readthedocs, but I think its fine to keep here for now. 

I'll add comments here, since its difficult to provide review comments on a notebook at the moment. 

- For the installation instructions you could look at how Sophie has set up the continuous integration using `wget` to download the packages and install them. We could either recommend that here, or at least provide the URLs for people in the notebook. 
- Step 0: Emphasize that you need to do this for every python file or notebook that you're running (i.e. it's not just a one-time thing). 
- Step 1: 
    - I'd say for now don't even mention a custom spectrum until we have that API better defined and a specific example just for that feature. 
    - When accessing the spectra you could mention that under the hood we use synphot, so that people can know how to manipulate the object. Also, highlight that `sp_comp` is a list. 
    - I believe the quantity you're plotting is actually the "Flux Density", or more specifically the "Photon Flux Density" (Flux density is fine I think).  
- Step 2: 
    - "Step2" -> "Step 2"
    - I think we can probably drop "lowfs". I don't expect to be using corgisim to simulate the low order wavefront sensor.
    - Can you put a link to cgisim_public_v4.0.pdf? 
    - Similarly, can you point towards where to find the DM solutions? 
    - For now you can leave out the `polaxis` option. Unless we're specifically working with the polarization mode we won't need it. It should still be accessible of course for a power-user via proper_keywords, but for basic usage we don't need to bring it up. Polarization just confuses people. 
    - What exactly does the `setup_bandpass` function do? It looks like its got the same inputs as the init function for CorgiOptics. Is it needed? Is it only here to grab the filter info to plot? If so, make that clear. If a user needs to run it themselves (they probably shouldn't have to), we just need to clarify that. What is the "0" in the input? 
    - It could be helpful to mention in the tutorial how things are developed with HDUs, so that the scene components are accessible with a data and header component.  
- Step 3: 
    - "Step 3, " -> "Step 3:"
- Step 4: 
    - "Step 4," -> "Step 4:"
    - For both this step and Step 2, it would be good to describe how the input keyword dictionaries work: that they are the keyword arguments to the internal functions for cgisim and emccd_detect, and that everything stays the default settings unless otherwise changed. Is there a way to access the default settings? To show in this tutorial? 
    - "we'll use the default parameters for the EMCCD detector (recommended)": maybe take away the "recommended" part. I'm not sure what we should recommend yet. 
- Really the final step in creating an L1 data product will be populating the headers in an output file. It's worth mentioning that this is a work in progress (but a critical step).

maxwellmb (changes requested):
Overall I think this is a good strategy for now as we bring people up to speed. In the not-too-distant future, we might split out some of the things like the install instructions to be their own pages on the readthedocs, but I think its fine to keep here for now. 

I'll add comments here, since its difficult to provide review comments on a notebook at the moment. 

- For the installation instructions you could look at how Sophie has set up the continuous integration using `wget` to download the packages and install them. We could either recommend that here, or at least provide the URLs for people in the notebook. 
- Step 0: Emphasize that you need to do this for every python file or notebook that you're running (i.e. it's not just a one-time thing). 
- Step 1: 
    - I'd say for now don't even mention a custom spectrum until we have that API better defined and a specific example just for that feature. 
    - When accessing the spectra you could mention that under the hood we use synphot, so that people can know how to manipulate the object. Also, highlight that `sp_comp` is a list. 
    - I believe the quantity you're plotting is actually the "Flux Density", or more specifically the "Photon Flux Density" (Flux density is fine I think).  
- Step 2: 
    - "Step2" -> "Step 2"
    - I think we can probably drop "lowfs". I don't expect to be using corgisim to simulate the low order wavefront sensor.
    - Can you put a link to cgisim_public_v4.0.pdf? 
    - Similarly, can you point towards where to find the DM solutions? 
    - For now you can leave out the `polaxis` option. Unless we're specifically working with the polarization mode we won't need it. It should still be accessible of course for a power-user via proper_keywords, but for basic usage we don't need to bring it up. Polarization just confuses people. 
    - What exactly does the `setup_bandpass` function do? It looks like its got the same inputs as the init function for CorgiOptics. Is it needed? Is it only here to grab the filter info to plot? If so, make that clear. If a user needs to run it themselves (they probably shouldn't have to), we just need to clarify that. What is the "0" in the input? 
    - It could be helpful to mention in the tutorial how things are developed with HDUs, so that the scene components are accessible with a data and header component.  
- Step 3: 
    - "Step 3, " -> "Step 3:"
- Step 4: 
    - "Step 4," -> "Step 4:"
    - For both this step and Step 2, it would be good to describe how the input keyword dictionaries work: that they are the keyword arguments to the internal functions for cgisim and emccd_detect, and that everything stays the default settings unless otherwise changed. Is there a way to access the default settings? To show in this tutorial? 
    - "we'll use the default parameters for the EMCCD detector (recommended)": maybe take away the "recommended" part. I'm not sure what we should recommend yet. 
- Really the final step in creating an L1 data product will be populating the headers in an output file. It's worth mentioning that this is a work in progress (but a critical step).

maxwellmb:
- `roman_preflight_proper.copy_here` here needs to be a called function: `roman_preflight_proper.copy_here()`
- The install instructions don't include the install instructions for requirements.txt. like in the tutorial. Doesn't pytest get installed by requirements.txt? If so, it probably doesn't need to be explicitly listed in the tutorial.
- You can remove 'lowfs' from the simulation mode options and you can remove 'zwfs' from the coronagraph types. 
- We still need to get rid of all the polaxis stuff for now. 
- It would be good to put some text right after the Step 3 title to explain the details of the next few lines, even if its obivous to us. It makes me wonder actually if we should rename the function, since get_psf is a bit ambiguous. maybe get_host_psf?

maxwellmb:
(but I can confirm that it all runs on my computer - after fixing the copy_here() thing)

maxwellmb:
I think the last thing missing is that you need to add the `pip install -r requirements.txt` instructions. Then we're good to go. Nice work!

Issue #31, opened 2025-04-10
Spectral type lower case


helloWorlds0:
The spectral types (M5V','M0V','K5V','K0V','G0V','F5V','B3V','A5V','A0V') are in upper case in the code but in lower case in the file names of cgisim_info_dir (g0v.dat). This works on Mac and linux, where name files are not case sensitive, but fails in Linux, where they are. 

The spectral types should either be in lower case everywhere or be made to be case insensitive

helloWorlds0:
After talking with Jingwen, the workaround is to keep uppercase in our code, but to use lowercase when we call cgisim for comparison in  testing

PR #35, opened 2025-04-15
Modify tests to comply with pytest

helloWorlds0:


helloWorlds0:
Changed the name of the test and made the plotting optional. I DID NOT made the changes to make the opening of the files case insensitive (issue #31) so the tests fail, but they are collected

helloWorlds0:
Re-merged main into it, fixed a few lower case bugs that had slipped through plus an issue with a type error in test off axis source spectrum

maxwellmb (approved):

PR #36, opened 2025-04-15
modify sptype to allow opening of cgi files on linux

helloWorlds0:


maxwellmb (approved):
Looks good, thanks for fixing this!

PR #41, opened 2025-04-21
add_header_info from drp

zjwhale:
1) add header
Now we read header from Corgidrp to pass the header info. Besides that, we add comments header to track simulation setup
2) add corgidrp to requirements


corgisim/instrument.py
```diff
@@ -598,45 +600,69 @@ def define_EMCCD(self, emccd_keywords=None ):
...
+        # Add descriptive comments to the primary header
+          
+            primary_hdu.header['COMMENT'] = "This FITS file contains simulated data."
+            primary_hdu.header['COMMENT'] = "Primary header includes stellar properties and simulation details."
+
+            for key, value in sim_info.items():
                 comment = key+' : '+str(value)
-                hdul.header.add_comment(comment)
-                #hdul.header[key] = value
+                primary_hdu.header.add_comment(comment)
```
maxwellmb, line 642: Have you thought about how to scrape this data back out in order to rebuild a simulation?

examples/exmple_on_axis_star_on_dector_fullframe.py
```diff
@@ -38,18 +39,20 @@ def test_on_axis_star_on_detector_fullframe():
...
 
     detector = instrument.CorgiDetector( emccd_keywords)
     #print('aaa',detector.emccd.meta.geom )
     sim_scene = detector.generate_detector_image(sim_scene, exptime,full_frame=True,loc_x=300, loc_y=300)
-    image2 = sim_scene.image_on_detector.data
+    image2 = sim_scene.image_on_detector[1].data
+
 
 
+    instrument.save_hdu_to_fits(sim_scene.image_on_detector, outdir='/Users/jingwenzhang/Desktop/corgisim/examples')
```
maxwellmb, line 55: This directory will likely cause problems on other people's computers. Try to save something more locally in the repository. A folder like test_data could work well. 

What is the point of saving this file?

maxwellmb (changes requested):
There are some small comments here. The bigger comment is that we now need a test to verify that the saving of the L1 product is correct. This would just open up the newly created file and cross-check that the expected header keywords are there, and that the format of the data is as expected.

corgisim/instrument.py
zjwhale: YES

examples/exmple_on_axis_star_on_dector_fullframe.py
maxwellmb: I'm going to resolve this comment, since we've decided that it's worth it. Right?

corgisim/instrument.py
```diff
@@ -598,45 +647,71 @@ def define_EMCCD(self, emccd_keywords=None ):
         return emccd
 
 
-def create_hdu(data, header_info=None):
+def create_hdu(data, sim_info=None, header_info = None, write_as_L1=False):
```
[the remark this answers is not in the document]
maxwellmb: Why are we creating hduls here, but saving them in scene.py? I think we should probably do them in the same place. Do we want a new .py file to do this in? Maybe io.py?

examples/exmple_on_axis_star_on_dector_fullframe.py
maxwellmb, line 1: was this file renamed on purpose?

test/conftest.py
```diff
@@ -0,0 +1,23 @@
+from corgisim import scene
+from corgisim import instrument
+import proper
+import roman_preflight_proper
+import pytest
+import shutil
+import os
+
+def pytest_sessionstart(session):
```
maxwellmb, line 9: I don't really understand what this file and function are supposed to be doing. When is this test supposed to be called? It doesn't look like it gets called by pytest and its unclear what its supposed to be testing. It looks like its copying some of the roman proper files, but its unclear to what end.

corgisim/instrument.py
```diff
@@ -386,24 +398,35 @@ def inject_point_sources(self, input_scene, sim_scene=None, on_the_fly=False):
...
+        keys_to_include_in_header = [ 'use_errors','polaxis','final_sampling_m', 'use_dm1','use_dm2','use_fpm',
+                            'use_lyot_stop','use_field_stop']  # Specify keys to include
         subset = {key: self.proper_keywords[key] for key in keys_to_include_in_header if key in self.proper_keywords}
-        header_info.update(subset)
+        sim_info.update(subset)
+        sim_info['includ_dectector_noise'] = 'False'
         # Create the HDU object with the generated header information
-        sim_scene.point_source_image = create_hdu(np.sum(point_source_image,axis=0),header_info =header_info)
+
+        sim_scene.point_source_image = create_hdu( np.sum(point_source_image,axis=0), sim_info =sim_info)
```
maxwellmb, line 429: it seems like point_source_image is now an hdu list rather than an hdu. Can you tell me the reason for this change? Obviously its related to the fact that we want the output fits file to be formatted in this way as an hdu list, but it's not immediately clear to me why we want the internal storage to be like that. I'm not fully against this, but I'd like to understand why.

examples/host_star_image.fits
maxwellmb, line 1: Does this file need to be in the git repository? Why was it added to this commit?

maxwellmb (changes requested):

corgisim/instrument.py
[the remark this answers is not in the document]
zjwhale: I think to create a separate io.py is a good idea. We can put both create_hdu() and save_hdu_to_fits() in there. But I suggest to use a different name other than io.py because there is a standard Python library called io as well.

examples/exmple_off_axis_source_on_dector_fullframe.py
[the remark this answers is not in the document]
zjwhale: It’s not a test file—it's a script I use to generate plots and check the output, so I placed it in the `examples` folder. The actual new test file is called` test_L1_product_fits_format.py` and is located in the `test` folder.

examples/CGI_0000000000000000000_20250426T1749252_L1_.fits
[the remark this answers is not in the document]
zjwhale: I included it so you can check the output FITS file. Where do you prefer I share the FITS file with you— Slack, or here?

examples/exmple_on_axis_star_on_dector_fullframe.py
zjwhale: I don’t consider this a test file since it doesn’t actually perform any tests—it’s just for plotting and checking outputs. So I renamed it from test_ to example_ and moved it to the examples folder. For this type of files, would you prefer me to exclude them from tracking and hide them from git pull?

test/conftest.py
zjwhale: I added a new test called test_L1_product_fits_format.py, which generates an on-sky L1 product FITS file and reads it back to validate the header. However, during integration testing, I encountered an error due to the missing Roman_preflight.py prescription file. Sophine pointed out that this might be because tests run independently in alphabetical order, and this test runs before test_installation.py, which normally sets things up.

To fix the issue, Jason suggested using a conftest.py file to configure the necessary setup before any tests run. I tried this approach and it worked well. I think it's a better practice, since it removes the dependency on test_installation.py.

A details about conftest.py shared by Jason is here: https://stackoverflow.com/questions/49685755/pytest-run-a-start-up-script

corgisim/instrument.py
zjwhale: The only reason is because I use the same `create_hdu()` function for `host_star_image`, `point_source_image` and `image_on _detector.` But if you prefer, I can also write a different function that saving intermediate images as a HDU.

examples/host_star_image.fits
zjwhale: Like the other FITS files, I’ve included them so you can review the output.

examples/CGI_0000000000000000000_20250426T1749252_L1_.fits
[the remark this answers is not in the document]
maxwellmb: what do you mean check it? Like as part of a test? Or just for me to take a look? If it's just for me then we shouldn't save it in the repository here.

examples/exmple_on_axis_star_on_dector_fullframe.py
maxwellmb: What exactly is it doing? Does it make sense to put things into tutorial notebooks, if there's something someone could learn about the corgisim infrastructure? 

If it's just "Jingwen's sanity check", then it doesn't have to be part of the official package.

test/conftest.py
maxwellmb: Ah right. Sorry, so this is that thing. Ok. It might help to add a few comment lines to explain this.

examples/CGI_0000000000000000000_20250426T1749252_L1_.fits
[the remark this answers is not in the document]
zjwhale: Yes, I can remove them then

examples/exmple_off_axis_source_on_dector_fullframe.py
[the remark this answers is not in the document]
maxwellmb: It's everything that's done in this file covered in the tutorial now?

examples/host_star_image.fits
maxwellmb: We can remove it then.

examples/exmple_off_axis_source_on_dector_fullframe.py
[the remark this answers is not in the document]
zjwhale: Yes, I will remove both `example_off_axis_source_on dector_fullframe.py` and `example_on_axis_star_on dector_fullframe.py`

test/test_L1_product_fits_format.py
```diff
@@ -0,0 +1,97 @@
...
+    cor_type = 'hlc_band1'
+
+    mag_companion = [25,25]
+    ###the position of companions in unit of mas
+    ####550nm/2.3m = 29.4 mas
+    ###we used sep = 3 lambda/D here 
+    dx= [3*49.3,-3*49.3]
+    dy= [3*49.3,-3*49.3]
+    
+    info_dir = cgisim.lib_dir + '/cgisim_info_dir/'
```
maxwellmb, line 32: What's this all about? info_dir doesn't get used

maxwellmb: If we don't use it, then we can get rid of the stuff from cgisim.lib_dir in __init__.py

maxwellmb (changes requested):
Super minor things

test/test_L1_product_fits_format.py
zjwhale: `info_dir = cgisim.lib_dir + '/cgisim_info_dir/' `is used to get the path to the cgisim installation directory on the user's local machine, since we need to read DM files from within the installed cgisim folder.

Similarly, I added a new line in __init__.py to enable retrieval of the local installation path using `corgisim.lib_dir`.  This allows us to access FITS files from the /test/testdata folder when running test scripts.

maxwellmb (approved):

PR #43, opened 2025-04-30
15 convert observing setup from cpgs to inputs for corgisim

helloWorlds0:


test/test_data_loading.py
maxwellmb, line 1: similar to the comment about, instead of 'data', let's be specific about 'cpgs'. In the future I imagine we'll want to expand these tests, but this is great for now.

corgisim/data_loader.py
maxwellmb, line 1: perhaps the name should be about 'cpgs' rather than data to be a bit less ambiguous.

corgisim/data/cpgs_default.xml
maxwellmb, line 1: Could we move this file and the next one into a `test/test_data/` folder?

corgisim/scene.py
```diff
@@ -320,7 +320,7 @@ def get_stellar_spectrum(self, sptype, magnitude, magtype = 'vegamag' ):
             # read vega spetrum
             vega_spec = SourceSpectrum.from_vega()
             # sp_scale has the same units as sp
-            sp_scale = sp.normalize(renorm_val= magnitude * VEGAMAG, band=v_band,vegaspec = vega_spec )
+            sp_scale = sp.normalize(renorm_val= magnitude * VEGAMAG, band=v_band,vegaspec = vega_spec, force = True )
```
maxwellmb, line 323: What does this do?

maxwellmb (changes requested):

corgisim/data/cpgs_default.xml
helloWorlds0: Yes

corgisim/data_loader.py
helloWorlds0: I think the _file_ should be about data (so that we put all functions that load data in there), but the function in itself is specifically about cpgs

test/test_data_loading.py
helloWorlds0: Same as above: I think the _file_ should be about data loading, even if the test is only about cpgs. In the case of tests specifically, I think we should reorganize the whole folder, since right now it's one file <-> one tests, even if they tests similar things

corgisim/scene.py
helloWorlds0: This is a bit of a weird one: In the synphot library ([here](https://synphot.readthedocs.io/en/latest/index.html), the normalize function returns an error if there is only a partial overlap between 'self' and v_band. V_band is defined by us as `SpectralElement.from_filter('johnson_v')`, but I don't know what this actually mean or why we choose it. self comes from our spectral type, so in this case the cpgs file. Putting `force = True` allows us to continue even with this error. 
Since this error isn't linked to CPGS, I chose to just push through, but I think it might need its own issue or at least an explanation

maxwellmb: @zjwhale, do you know what's up here? If its throwing an error we may need to look more closely at this.

test/test_data/CGI_0000000000000000014_20221004T2359351_L1_.fits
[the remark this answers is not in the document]
maxwellmb: if its generated by a test, perhaps we can add some clean-up code. Or we just don't track it.

maxwellmb (changes requested):

corgisim/data_loader.py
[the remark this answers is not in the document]
helloWorlds0: We mentioned it briefly during last week meeting; it would depend on user cases, but it could be json files, user input, or basically any input (command line argument, etc.)

test/test_data/CGI_0000000000000000014_20221004T2359351_L1_.fits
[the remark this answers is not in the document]
helloWorlds0: We need it to test the data loading (there are two cpgs files and a fits file which is used to test that the correct error is raised when the input file is not an XML file)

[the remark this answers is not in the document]
maxwellmb: Can't we just feed it a .py filename or something?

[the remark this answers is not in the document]
helloWorlds0: yes, we can give it basically anything, I used the fits file because it was there

corgisim/data_loader.py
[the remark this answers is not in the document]
maxwellmb: Alright, well it's fine to have a general input file like this, but "data" isn't the right word here. We'll have to brainstorm another. Maybe configuration_loader? I'm open to suggestions.

test/test_data/CGI_0000000000000000014_20221004T2359351_L1_.fits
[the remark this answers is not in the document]
maxwellmb: I don't think it is anymore.

corgisim/data_loader.py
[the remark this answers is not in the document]
helloWorlds0: then it could be inputs ? Whichever we choose, I don't think this is where we should change it.

test/test_data/CGI_0000000000000000014_20221004T2359351_L1_.fits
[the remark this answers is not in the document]
maxwellmb: I doesn't look like this was resolved. Please don't add this file back in.

corgisim/data_loader.py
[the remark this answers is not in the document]
maxwellmb: Sure we can do `inputs.py`. What do you mean you don't think this is where we should change it?

corgisim/scene.py
zjwhale: @helloWorlds0 The V band is part of the standard Johnson photometric system and corresponds to a specific range of visible wavelengths used to measure stellar brightness. In CorgiSim, we use the stellar V-band magnitude as an input, which is why we set: `Vband = SpectralElement.from_filter('johnson_v')`

TThe code is intended to scale the stellar spectrum based on the input V-band magnitude to ensure the correct flux intensity. The error likely occurs because the initial spectrum: `sp = SourceSpectrum(BlackBodyNorm1D, temperature=v0) ` has a wavelength range that does not overlap with the V-band filter. This is unexpected, as most stellar spectra should naturally cover the V band.

Therefore, rather than setting force=True, it would be better to investigate why the initial spectrum lacks overlap with V-band in the first place.

maxwellmb: Ok, @zjwhale once the issue is created, then I'll resolve this comment so that we don't lose track of it.

zjwhale: I’ve created a new PR (#51) to fix the bug in the spectral type–temperature mapping function. Specifically, the code now automatically appends "V" if no luminosity class is given.

corgisim/data_loader.py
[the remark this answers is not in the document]
helloWorlds0: I mean in this issue

corgisim/scene.py
helloWorlds0: Why "V" instead of another class ?

helloWorlds0: More precisely, could you point me to some resources concerning specrtal type? Based on thecode right now, the temperature is only defined by the rank, which means that it's defined by the function `sort_sptype` . So the rank is the same, for instance, for G1I, G1IV, G1II, or G1I.

test/test_data/CGI_0000000000000000014_20221004T2359351_L1_.fits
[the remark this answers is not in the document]
helloWorlds0: Done !

corgisim/data_loader.py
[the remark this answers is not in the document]
maxwellmb: Ok....I don't see why we wouldn't rename it now, before other people start working on it and the name propagates to other places.

corgisim/scene.py
maxwellmb: The spectra are basically the same for any luminosity class, so it probably doesn't really matter too much. Most stars will be main sequence stars "V", so that's a fine decision in most cases. The wikipedia article is a reasonable summary: https://en.wikipedia.org/wiki/Stellar_classification#Luminosity_class

corgisim/data_loader.py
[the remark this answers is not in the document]
helloWorlds0: I can rename it now, if we chose to go with inputs, but if I continue with cpgs I can also change it in the next issue. But I think it would be more efficient to let the code go into the main branch and create another issue to discuss the name than discuss it in the peer review

PR #45, opened 2025-05-02
rename class SimulatedScene to  SimulatedImage

zjwhale:


.gitignore
maxwellmb, line 1: something is wrong here, we shouldn't have all this in the .gitignore

examples/tutorial1_corgisim.ipynb
```diff
@@ -579,14 +579,14 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 13,
+   "execution_count": 14,
    "metadata": {},
    "outputs": [],
    "source": [
     "## Pass the base_scene object to corgi.optics and use get_psf to simulate the host star PSF.\n",
     "## The result is stored in a SimulatedScene object as an Astropy HDU containing both data and header information.\n",
```
maxwellmb, line 587: I think this line needs updating

maxwellmb (changes requested):

maxwellmb:
There are also some merge conflicts. Please resolve them!

examples/tutorial1_corgisim.ipynb
zjwhale: This line indicates the execution order of the code cell in the Jupyter notebook. The change from 13 to 14 occurred because a new code cell was added before it. What updates are needed here?

.gitignore
zjwhale: Ah, I thought everything we didn’t want them in the repo should go in .gitignore. I’ll  remove those files.

zjwhale:
> There are also some merge conflicts. Please resolve them!
I did not see any merge conflicts in this page. Could you point out what are they?

examples/tutorial1_corgisim.ipynb
maxwellmb: I was commenting here about the line in the comment "The result is stored in a SimulatedScence Object", should be SimulatedImage now, right?

maxwellmb: I'm going to let this through to move on though

maxwellmb (approved):

PR #46, opened 2025-05-02
add_code_when_imagesize _is_1024

zjwhale:
Previously, we only handled the case where the noise-free image generated by corgi.optics was smaller than 1024×1024, requiring it to be placed on the detector using place_scene_on_detector(). In this branch, I’ve added support for the case where the image is exactly 1024×1024, allowing us to skip the placement step entirely.


maxwellmb (commented):
This looks fine after the other two PRs are merged in.

maxwellmb (changes requested):
This looks fine after the other PRs are merged, with one exception: we should introduce a new test to make sure this is working correctly.

zjwhale:
> This looks fine after the other PRs are merged, with one exception: we should introduce a new test to make sure this is working correctly.

Yes, I added a testing module about the new part in `test_off_axis_source.py`

maxwellmb (approved):

maxwellmb:
@zjwhale this is good to go, except for merge conflicts with the tutorial notebook. It could be a good chance to update the SimulatedImage that I mentioned in the other PR.

PR #50, opened 2025-05-09
add empty scene

zjwhale:
Add a empty_scene to the SimulatedScene() class. It is for dark and bias simulation

### DIFF
--- a/corgisim/instrument.py
+++ b/corgisim/instrument.py
@@ -431,6 +431,39 @@ class CorgiOptics():
 
         return sim_scene
 
+    def generate_empty_scene(self, input_scene, sim_scene=None):
+        """
+        Generate an empty scene to simulate a bias or dark frame.
+
+        Arguments
+        ----------
+        sim_scene : corgisim.SimulatedScene, optional
+            An existing SimulatedScene object to attach the empty image to. 
+            If None, a new SimulatedScene will be created.
+
+        Returns
+        -------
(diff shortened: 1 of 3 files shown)


maxwellmb (changes requested):
I don't really understand this implementation. Why do we want it as a new property in a scene? When would we use that property other than just creating an empty scene? When generating an empty scene wouldn't we just set total_image to be zeros? 

I also don't understand why optics would be generating it. It seems to have nothing to do with the optics. 

Finally, I think we'd like to test that the final sim_scene has properties that we'd expect (e.g. the expected read noise to within a tolerance).

maxwellmb:
From Jingwen and Max's chat: 

Changes to make: 

- in `CorgiDetector.generate_dtector_image()` move the scene combination to `SimulagedImage.combin_simulated_scene_list()` (consider renaming to just `combine_simuated_scene()`. This will populate SimulatedImage.total_image. 
- CorgiDetector.generate_detector_image() will by default operate on SimulatedImage.total_image. If it is None, then (when a flag is set) it will call combine_simulated_scene() itself. 
- We can define an EmptyScene as a child class of SimulatedScene, as below (or something like it). 

```
class EmptyScene(SimulatedScene):
    def __init__(self,size=1024)
        super().__init__(None) #We don't have an input scene to initialize, so we pass None. 
        
        self.total_image = np.zeroes((size, size))
        ```

PR #51, opened 2025-05-09
fix_bug_sptype_teff_mapping

zjwhale:
This branch is to fix the biug in spectral type and temperature mapping function. Specifically, the code will auto-append "V" if no luminosity class is given.


maxwellmb (changes requested):
Can we include this in very simple a test somewhere?

zjwhale:
> Can we include this in very simple a test somewhere?

Unlike what we discussed this afternoon (the code only handled luminosity class V), I’ve now fixed the spectral type to effective temperature mapping. The updated code supports luminosity classes I, II, III, IV, and V. I’ve also added a test file named `test_sptype_teff_mapping.py.`

maxwellmb (approved):

PR #60, opened 2025-05-28
Update python-app.yml

helloWorlds0:


helloWorlds0:
Downgraded to using ubuntu 22.04

PR #64, opened 2025-06-12
 Add Field-Dependent 2D Convolution

### LINKED ISSUE #17: Add 2D convolution
Adding 2D scene, such as disk, in the Scene Class. There are some few pre-existing codes in example.py file.

alexisyslau:
### Update (2025-02-26) - Ready for review
- Now`simulate_2d_scene` has been moved into the `CorgiOptics` following the definition of `inject_point_sources()` and `get_host_psf()`, with `prf_cube_path` as an input.
- Updated the notebook to reflect the lastest changes 
- Rebased onto the main 

### Update (2025-01-07) 

1. Module Restructuring: 
- Created `prf_simulation.py` : separate PRF generation logic from `instrument.py`, as `instrument.py` should only contain the optical setup
  - Moved `make_prf_cube()` , `compute_single_off_axis_psf()`-> `prf_simulation.py`
 
- Refactored `convolution.py`:
 - moved the 2D_scene convolution from `instrument.py` to `convolution.py`
 - flux calibration and `sim_info` are separated from the convolution logic 

2. Improved position handling
-  Excluding on-axis PSF (all positions at r=0)

3. PRF Cube and metadata handling
- PRF cube saved as fits file with descriptive naming: ` prf_cube_{cgi_mode}_{cor_type}_band_{bandpass}.fits`
- Simulation parameters stored in the FITS COMMENT `(sim_info)` header. 
 - These parameters will be extracted during the convolution process

4. Updated demos: 
- Simplified `corgisim_convolution.ipynb`: 
 - `disk_demo.ipynb`: disk convolution with pre-computed PRF cube
 - `prf_cube_build_demo.ipynb`: build the prf_cube

5. Updated tests: updated `test_2D_scene.py` 

6. Added PRF centring when the input PRF cube is not centred.
Convolution assumes the PSF/PRF is defined relative to the source (i.e. its peak is at zero offset). Off-axis PRFs are typically stored with a displaced peak; using them directly would apply that displacement again when the PRF is placed during convolution, causing an unphysical double shift. Re-centring the PRF ensures the offset is applied exactly once.

### Previous updates
<details>
<summary>click to expand</summary>

### Update (2025-12-15) 
- TODO: Flux injection and calibration:  @ChenXie-astro (new PR)?

</details>

### Original PR description (outdated)
<details>
<summary>click to expand</summary>

## Purpose
Introduce a minimal-viable 2-D field-dependent convolution (resolve #17)
(shift–add with either nearest-neighbour or bilinear PRF selection) that we can migrate into the core CorgiSim framework.

`example/corgisim_convolution.ipynb` will show the step-by-step guide on how to use the convolution functions, and detailed explanations.  

##  Summary of Changes
- **Major refactor of** `convolution.py`
- Added unit tests and error handlings for functionalities in `convolution.py` and cases handling in `instrument.py`
- Updated `scene.py` in how to handle `twoD_scene_info`.

Multiple helper functions are created for:
- padding PRF kernels 
- building sampling grids, and 
- calculating weights for convolution _(for optional bilinear method, which uses fractional pixel offsets within the radial-azimuth cell)._
- binning images (may need this when you are performing convolution, depends)
- resize_array (may need this when you are performing convolution, depends)
- `_compute_single_off_axis_psf` which takes `lambda / D , theta ` as input
- **(not live yet)** Fourier shift of the PRF before convolution so the core of the PRF is in the centre of the frame before making convolution

Example: create a **small PRF cube** with a toy grid 

```

radii\_lamD\_1 = conv.build\_radial\_grid(
inner\_step      = 5,     # step inside IWA (λ/D)
mid\_step        = 5,     # step between IWA and OWA
outer\_step      = 5,     # step beyond OWA (λ/D)
iwa             = 0,     # inner working angle (λ/D)
owa             = 10,    # outer working angle (λ/D)
)

azimuths\_deg\_1 = conv.build\_azimuth\_grid(
step\_deg        = 90  # azimuthal step in degrees
)

# making the cube

prf\_cube = optics.make\_prf\_cube(
radii\_lamD\_1,
azimuths\_deg\_1)

```

_No fractional offsets (nearest neighbour)_ 
```

# without bilinear interpolation

sim\_scene = optics.convolve\_2d\_scene(input\_scene=base\_scene, sim\_scene=sim\_scene, use\_bilinear\_interpolation=False)
image\_disk\_corgi = sim\_scene.twoD\_image.data

```
image

_With fractional offsets (bilinear weighting)_
```

# bilinear weighting (smooth)

sim\_scene = optics.convolve\_2d\_scene(input\_scene=base\_scene, sim\_scene=sim\_scene, use\_bilinear\_interpolation=True)
image\_disk\_corgi = sim\_scene.twoD\_image.data

```
image

**Illustration of how does shift-add convolution is performed** 
build_up_smooth

##  Status

- [x] Initial implementation
- [x] Code cleanup and optimization
- [ ] Full test coverage
- [ ] Example notebook / demo script _(update is needed; 1: convolution + flux calibration 2: making the prf cube)_

## What Needs Review
- Logic of our field-dependent convolution approach, especially bilinear weighting (smooth)
- Compatibility with CorgiSim (e.g. `scene`, `instrument`)

##  Testing
- Manual check with synthetic inputs, e.g. 51 × 51 synthetic scene (point source + disk) blurred with a 9-PRF grid. (Done)
- Port local tests to `test_2dscene.py` (Done) 

##  Known Issues / Notes
- Flux Calibration _(not in the scope of this PR, @ChenXie-astro should open a new PR)_
- _compute_off_axis_psf does not multiply with counts, we perform convolution first 
- make_prf_cube (source_sed --> it is set to flat). At the moment, not sure whether it should be flat, disk spectrum (or the star) #140 
</details>


corgisim/convolution.py
```diff
@@ -3,289 +3,277 @@
...
+    t0 = time.time(); _bar(0)
+    for idx, (r_ld, th_deg) in enumerate(zip(rr, tt), 1):
+        rad = th_deg.to_value(u.rad)
+        run_opts = {**prop_opts,
+                    "source_x_offset": r_ld * np.cos(rad),
+                    "source_y_offset": r_ld * np.sin(rad)}
+
+        wfs, _ = proper.prop_run_multi(prop_tag, lam_grid,
+                                       prf_width_px, QUIET=True,
+                                       PASSVALUE=run_opts)
```
zjwhale, line 215: In `CorgiSim`, we wrap around `cgisim`, which internally calls `PROPER` for simulating Roman PSFs. It is based on a configuration file called `roman_preflight`, which defines the CGI optical settings. The major comment is to revise how `PROPER` is called here. 

Details about cgisim and roman_preflight can be found: https://sourceforge.net/projects/cgisim/ 

Some example used in corgisim like:

(fields, sampling) = proper.prop_run_multi('roman_preflight',  self.lam_um, 1024,PASSVALUE=self.proper_keywords,QUIET=self.quiet)

alexisyslau:
Need to fix a few bugs before it is ready to be review again, in particular making sure that it works with both point_source and 2D scene.

corgisim/instrument.py
```diff
@@ -465,32 +475,286 @@ def setup_bandpass(self, cgimode, bandpass_name, nd ):
...
+            Azimuthal angles in degrees.
+        source_sed : array-like or None
+            Spectral energy distribution weights, optional.
+
+        Returns
+        -------
+        prf_cube : ndarray, shape (N_positions, Ny, Nx)
+            Cube of PSFs at all requested (r, θ) positions.
+        """
+        # TODO - source_sed is currently not used, just assumed to be flat. For scene with extended source, here should be disk spectrum?
```
maxwellmb: Can we make an Issue about this so the TODO doesn't get lost in the depths of the code?

corgisim/instrument.py
```diff
@@ -465,32 +475,286 @@ def setup_bandpass(self, cgimode, bandpass_name, nd ):
...
+        # Prepare additional information to be added as COMMENT headers in the primary HDU.
+        # These are different from the default L1 headers, but extra comments that are used to track simulation-specific details.
+        sim_info = {'ref_flag': False,
+                    'cgi_mode':self.cgi_mode,
+                    'cor_type': self.optics_keywords['cor_type'],
+                    'bandpass':self.bandpass_header,
+                    'over_sampling_factor':self.oversampling_factor,
+                    'return_oversample': self.return_oversample,
+                    'output_dim': self.optics_keywords['output_dim'],
+                    'nd_filter':self.nd}
```
maxwellmb: Can we add the filename to the PRF cube so that we can always keep track of what was done?

corgisim/convolution.py
```diff
@@ -3,289 +3,530 @@
...
-    theta_value = theta.to(u.deg).value  # Convert theta to degrees for comparison
-    kth = np.argmin(np.abs(thetas_value - theta_value))
-    theta_diff = theta - thetas[kth]  # Difference in angles
-
-    if kr == 0:
-        kpsf = 0
+import sys
+
+ARCSEC_PER_RAD = 206265
+PIXEL_SCALE_ARCSEC = 0.0218   # arcsec/pixel
```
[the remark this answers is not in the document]
zjwhale: The number looks good. It is the same as that in IPAC website: https://roman.ipac.caltech.edu/page/param-db#coronagraph_mode. But I think we should check with John that's consistent with what CGISim generated. I will ping him in the slack

test/test_2D_scene.py
maxwellmb: I'll come back to the tests after some of the above changes have been complete.

maxwellmb (changes requested):
This is an excellent start. A few comments related to the broader infrastructure:
- Let's make the disk input in physical surface brightness units, rather than relative to a star. We can have utility functions that converts a disk in contrast units into surface brightness if we'd like.
- How long does it take to generate a PRF cube? We should include how to generate it and a note about how long it takes (even if the code is commented out) in the tutorial.
- It would be helpful if the tutorial was actually run, showing output images.
- In order to simplify the storage and passing around of PRF cubes let's define a fits format for a PRF cube that includes the radii and azimuths for each cube stored in the same file in a different extension. That way we can just pass in one filename, or a hdulist to `convolve_2d_scene` with all the PRF contained in one place. This way we can instead keep the main input as a Scene and use scene._twoD_scene_hdu (a never a  SimulatedScene), and not have to include PRF information as part of a "scene". In general we're trying to keep scientific and instrument-related information separate. 

Let me know if you have questions about any specific comments!

corgisim/instrument.py
alexisyslau: Done. I will remove this line from the code

ChenXie-astro:
I haven’t uploaded my latest version of the flux calibration yet. It will be a completely separate function to handle the disk flux units, so you can remove the current code that performs the disk contrast scaling. I will create a new PR after your 2D PSF convolution is merged into main, which should keep things as clear as possible.

corgisim/instrument.py
```diff
@@ -465,32 +475,286 @@ def setup_bandpass(self, cgimode, bandpass_name, nd ):
...
+        return prf_cube
+
+    def convolve_2d_scene(self, input_scene, sim_scene=None, **kwargs):
+        """
+        Simulate a two-dimensional scene through the current CGI configuration.
+
+        This method retrieves the image data from `scene.background_scene` (or from
+        `scene.twoD_image` when `scene` is a `SimulatedImage`), then convolves it
+        with a grid of off-axis PSFs (also termed PRFs) either generated on the fly
+        or supplied via `prf_cube`. The convolution is flux-conserving and the result
```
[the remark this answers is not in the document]
alexisyslau: This logic is fixed now (if I understand correctly). Now we use `input_scene.twoD_scene_info` to handle the prf cube and disk model.

corgisim/instrument.py
alexisyslau: Done. Please check `mark_prf_cube` in `prf_simulation.py`

corgisim/convolution.py
```diff
@@ -3,289 +3,658 @@
...
+from synphot import units, SourceSpectrum, SpectralElement, Observation
+from corgisim import outputs, spec, prf_simulation, constants
+from corgisim.scene import SimulatedImage
+from scipy import interpolate
+import astropy.units as u
 
-    #Interpolate PRFs
-    prfs = fits.getdata(save_path)  # Load the original PRFs
-    interpolated_psfs_array = np.zeros_like(psfs_array)
+def binning(img: np.ndarray, binning_factor: int) -> np.ndarray:
```
maxwellmb: we're not really doing this kind of typecasting anywhere else, so I don't think we need to do it here.

corgisim/convolution.py
```diff
@@ -3,289 +3,658 @@
...
+
+    # Sum each bin x bin subarray
+    binim = np.sum(binim, axis=3)
+    binim = np.sum(binim, axis=1)
+
+    return binim
+
+def resize_array (array: np.ndarray, size:int, cent = None) -> np.ndarray:
+    """
+    Resize the array to a given size. "cent" is optional.
```
maxwellmb: Can you move the dosctring more decriptive? There are many ways to resize an array.

corgisim/convolution.py
```diff
@@ -3,289 +3,658 @@
...
-    ax.set_ylabel('arcsec')
-    ax.set_xlabel('arcsec')
-    fig.colorbar(im, orientation='vertical')
-    plt.show()
\ No newline at end of file
+    # input disk model
+    disk_model_data = fits.getdata(input_scene.twoD_scene_info['disk_model_path'])
+    disk_model_norm = disk_model_data/np.nansum(disk_model_data, axis=(0,1)) # normalisation of the disk
+
+    prf_cube_path = input_scene.twoD_scene_info['prf_path']
```
maxwellmb: We want to de-couple astrophysical things (contained in the Scene class) with instrument related things until we arrive at a Simulated Scene. Since prf_path isn't astrophysical and is instead instrument related, then it shouldn't be part of a Scene class. I think you could just have prf_path be in an argument to this function.

maxwellmb (approved):
In the interest of moving this ahead I wanted to send the current state of the review. I still haven't looked at the tests yet or closely at the notebook, but thought I'd send the current comment first.

maxwellmb (changes requested):
I didn't mean to approve it...

corgisim/convolution.py
```diff
@@ -3,289 +3,462 @@
...
-    ax.add_patch(circ1)
-    ax.add_patch(circ2)
-    ax.set_ylabel('arcsec')
-    ax.set_xlabel('arcsec')
-    fig.colorbar(im, orientation='vertical')
-    plt.show()
\ No newline at end of file
+    return conv
+
+def flux_calibration_2D_scene(optics, input_scene, conv2d):
```
maxwellmb, line 483: Could you please add a docstring here?

corgisim/convolution.py
```diff
@@ -3,289 +3,462 @@
...
+    yy, xx = np.indices(img_shape)
+    
+    radii_pixels  = np.hypot(xx - cx, yy - cy)
+    radii_lamD = (radii_pixels * pix_scale_mas) / res_mas
+
+    azimuth_deg  = np.degrees(np.arctan2(yy - cy, xx - cx)) % 360
+
+    return radii_lamD, azimuth_deg
+
+def _set_2D_image_sim_info(optics, input_scene):
```
maxwellmb, line 299: Could you please add a docstring here?

corgisim/prf_simulation.py
```diff
@@ -0,0 +1,389 @@
...
+        output_dir = './'
+
+    prf_cube_hdu.writeto(os.path.join(output_dir, prf_fname), overwrite=overwrite)
+
+    # TODO - maybe add a warning if overwriting existing file?
+    print(f"PRF cube is saved to {os.path.join(output_dir, prf_fname)}")
+    
+    return prf_cube_hdu
+
+def fourier_shift(img, shift):
```
maxwellmb, line 325: Could you please add a docstring here?

corgisim/scene.py
```diff
@@ -123,8 +122,22 @@ def __init__(self, host_star_properties=None, point_source_info=None, twoD_scene
                 pol.check_stokes_vector_validity(self.point_source_pol_state[source])
                 self.point_source_pol_state[source] = np.divide(self.point_source_pol_state[source], self.point_source_pol_state[source][0])
 
-        
-        
+        # setting up the 2D scene
+        # twoD_scene_info: a dictionary containing 2D scene information
+        self.twoD_scene_info = twoD_scene_info
+        self.twoD_scene_spectrum = None
+        self.twoD_prf_cubes = None
```
maxwellmb: prf cubes shouldn't be part of a scene since they're optics related. The Scene class is all astrophysical.

corgisim/instrument.py
```diff
@@ -1277,6 +1261,124 @@ def generate_full_aberration_psf(self, optics_keywords):
...
+
+        Returns
+        -------
+        sim_scene : SimulatedImage
+            Output scene with `twoD_image` replaced by the convolved and scaled result.
+
+        Raises
+        ------
+        ValueError
+            If `prf_cube_path` is missing or None.
```
maxwellmb: if it's a required argument then we shouldn't make it a keyword argument, but just an argument instead.

corgisim/instrument.py
```diff
@@ -1277,6 +1261,124 @@ def generate_full_aberration_psf(self, optics_keywords):
...
+        Notes
+        -----
+        - The PRF cube is assumed to be already centred in its arrays. 
+        - The PRF cube is assumed to be normalised to unit input flux. Absolute flux
+            scaling is applied in this function using the input scene spectrum and the
+            optics bandpass.
+        - The disk model is normalised by its total flux prior to convolution.
+        - After convolution, the image is scaled using the integrated bandpass count
+            rate and converted to a per resolution element normalisation using an
+            estimate of the PSF FWHM area (pixels) and a thresholded disk region mask.
```
maxwellmb, line 1453: Just to confirm, this is the part that's changing after @ChenXie-astro's upgrades?

corgisim/instrument.py
```diff
@@ -1277,6 +1261,124 @@ def generate_full_aberration_psf(self, optics_keywords):
...
+        # input disk model
+        disk_model_data = fits.getdata(input_scene.twoD_scene_info['disk_model_path'])
+        disk_model_norm = disk_model_data/np.nansum(disk_model_data, axis=(0,1)) # normalisation of the disk
+
+        prf_sim_info = prf_simulation._get_prf_sim_info(prf_cube_path) # Get the simulation information 
+
+        # Check if PRF cube needs centering
+        prf_info_is_centred = prf_sim_info.get('centred') # 'True' or 'False'
+        # Convert string to boolean
+        is_centred = (prf_info_is_centred == 'True')
```
maxwellmb, line 1478: Can you make an issue that it would be a nice-to-have feature to make sure that the current instrument configuration is the same as what is in the PRF?

corgisim/scene.py
```diff
@@ -13,7 +13,7 @@ class Scene():
     - Information about the host star (brightness, spectral type, etc.)
     - A list of point sources (brightness, location, spectra?, etc.)
     - A 2D background scene that will be convolved with the off-axis PSFs
-        - Format needs to be determined. Likely a fits hdu with specific header keywords. 
+        - Dictionary contains info required for convolution
```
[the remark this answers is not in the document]
maxwellmb: Unresolving this comment since at least the docs still say Dictionary. 

Also, please add some info in the Arguments here about what _needs_ to be in the hdu, or specifics about assumptions made (e.g. if the pixel scale is fixed, or based on a header keyword).

test/test_2D_scene.py
```diff
@@ -0,0 +1,341 @@
...
+    
+    # Test the key bilinear property: weights sum to 1
+    total_weight = sum(w[0, 0] for w in weights)
+    assert np.isclose(total_weight, 1.0)
+    
+    # Test that all indices are valid
+    all_indices = [idx[0, 0] for idx in indices]
+    assert all(idx >= 0 for idx in all_indices)
+
+def test_convolve_with_prfs_basic():
```
maxwellmb, line 194: Can we add a basic test of functionality somewhere? 

For example, if we have a 2D scene with just one pixel illuminated, it should be equal in brightness to if we just had a point source.

maxwellmb (commented):
Almost there I think! A few more comments (here and in-line): 
- How long does it take to generate a PRF cube? We should include how long it takes in the PRF tutorial. 
- It would be helpful if the tutorials were actually run, showing output images. 
- In order to simplify the storage and passing around of PRF cubes let's define a fits format for a PRF cube that includes the radii and azimuths for each cube stored in the same file in a different extension. That way we can just pass in one filename, or a hdulist to `convolve_2d_scene` with all the PRF contained in one place. This way we can instead keep the main input as a Scene and use scene._twoD_scene_hdu (a never a  SimulatedScene)
- It would be good to define point response functions in the tutorials. In the disk tutorial you should point people towards the PRF tutorial. Even just saying "PRF is another term for off-axis PSF when the PSF is field-dependent" or something like that. 
- What do you think of naming the tutorials "2D Scene (Disk) Tutorial 1: PRF Generation", "2D Scene (Disk) Tutorial 2: Image Generation" or something like that. 
- In the PRF tutorial can you just show how to save it (even though it's very simple). 
- Let's no advertise that this is fully up and running until we've applied the flux scaling fix

corgisim/convolution.py
alexisyslau: Done

corgisim/prf_simulation.py
alexisyslau: Done

corgisim/convolution.py
alexisyslau: Done

test/test_2D_scene.py
alexisyslau: Issue created  and I will address this in another PR

corgisim/instrument.py
alexisyslau: Done, issue created.

corgisim/instrument.py
alexisyslau: Yes.

corgisim/scene.py
```diff
@@ -59,15 +58,29 @@ class Scene():
...
-            pixel_scale, etc.
+        twoD_scene_info (dict): A dictionary containing information about an extended 2D scene, such as a disk or other extended structure. Required keys are:
+
+            - "contrast" (float): The contrast of the 2D scene relative to the host star in magnitudes (dMag). This value is currently used to generate a spectrum for the 2D scene based on the host star's spectrum, but in the future we may want to allow users to directly input the 2D scene spectrum or flux instead of relying on contrast with the host star.
+
+            - "disk_model_path" (str): The absolute file path to a FITS file containing the 2D scene image. The image data are treated as a spatial brightness template and are normalised internally so that the total scene flux sums to 1 before convolution. 
+
+                Assumptions:
+                - The FITS image must already be sampled on the same pixel scale used for convolution, currently 0.0218 arcsec/pixel, but in the future we may want to allow users to input the pixel scale as well.
+                - The image should be centred
```
maxwellmb, line 69: does the array size matter?

corgisim/scene.py
```diff
@@ -59,15 +58,29 @@ class Scene():
...
+                - The image should be oriented in the same reference frame as the point source coordinates (typically North-up, East-left).
+                - The current implementation does not resample the input image, or infer the pixel scale from the FITS header. 
+            
+            - "prf_cube_path" (str): The absolute file path to a FITS file containing the precomputed      off-axis PRF cube used to convolve the 2D scene.
```
maxwellmb: I don't think this needs to be here.

corgisim/instrument.py
```diff
@@ -1371,6 +1355,124 @@ def roll_angle(self, value):
...
+
+        # NOTE: An attempt to convert flux units to physical units after convolution
+        # NOTE: Should be remove in the future after we have a better way to track the units and perform the flux calibration in a more self-consistent way.
+        # 4. Flux calibration
+        flux_calibrated_conv2D = conv.flux_calibration_2D_scene(self, input_scene, conv2d)
+
+        if self.cgi_mode in ['spec', 'lowfs', 'excam_efield']:
+            warnings.warn(f"This mode '{self.cgi_mode}' has not implmented yet!") # still allow the usage but warn the user about this
+
+        sim_info = conv._set_2D_image_sim_info(self, input_scene)
```
maxwellmb, line 1519: is the PRF path saved in sim_info at all? We definitely want that for book keeping.

corgisim/instrument.py
```diff
@@ -1371,6 +1355,124 @@ def roll_angle(self, value):
...
+
+        if not is_centred:
+            print("PRF cube is not centred. centring now...")
+            from corgisim.prf_simulation import centre_prf_cube
+            prf_cube = centre_prf_cube(fits.getdata(prf_cube_path), method='centroid')
+        else: 
+            prf_cube = fits.getdata(prf_cube_path)
+
+        # Store the PRF cube in the sim_scene for potential later use (e.g., for visualisation or debugging)
+        sim_scene.twoD_prf_cubes = prf_cube
```
maxwellmb: unless we have a specific use-case at the moment, let's not carry this around "just in case". Instead let's just make sure to attach the filepath (see below).

corgisim/scene.py
```diff
@@ -546,6 +569,7 @@ def __init__(self, input_scene):
         #self.host_star_image_on_detector = None
         self.point_source_image = None
         self.twoD_image = None
+        self.twoD_prf_cubes = None # the precomputed PRF cube used to convolve the 2D scene. Saving for debugging purposes, but may not need to be stored? (memory intensive?)
```
maxwellmb: see comment elsewhere. We don't need to keep this here unless we have a specific use-case.

corgisim/instrument.py
```diff
@@ -1371,6 +1355,124 @@ def roll_angle(self, value):
...
+            prf_sim_info['inner_step'], 
+            prf_sim_info['mid_step'], 
+            prf_sim_info['outer_step'],
+            prf_sim_info['max_radius']
+        )
+
+        # 2. Get the azimuth grid for convolution
+        azimuths_deg, _ = conv.build_azimuth_grid(prf_sim_info['step_deg'])
+
+        # 3. Perform convolution
```
maxwellmb, line 1500: I think we need to make an issue that rolls the Scene given the telescope angle since that features is missing at the moment but is implemented for point sources.

test/test_2D_scene.py
maxwellmb, line 1: Do we test wide-field of view anywhere? It looks like it's an option in the tutorial, but it's not clear to me that it has been tested.

corgisim/instrument.py
maxwellmb: In the mean time we should throw an error if the roll angle isn't 0.

maxwellmb: (or feel free to introduce a roll).

maxwellmb (changes requested):
A small handful more!

corgisim/instrument.py
alexisyslau: I will introduce a roll but in another PR. Let's merge 2D convolution in first?
In the meantime I will add an exception if the roll angle isn't 0 
(added NotImplementedError, opened an issue)

corgisim/instrument.py
alexisyslau: Not at the moment. Added a line to do that in _set_2D_image_sim_info to get all input twoD_scene_info from the input_scene

corgisim/instrument.py
alexisyslau: Agreed. After checking the comment below, we can remove this line as user can access twoD_scene_info (e.g. prf path, disk path, unit e.t.c.) from the header of the output hdu. (Done)

corgisim/scene.py
alexisyslau: `- "prf_cube_path" (str): The absolute file path to a FITS file containing the precomputed  off-axis PRF cube used to convolve the 2D scene. ` : Yes

As for the object orientation and the pixel scale of the input object, I need to think about it. If not included in here, where should we put this information? Or this is not important at all?

corgisim/scene.py
alexisyslau: Agreed. I will remove this. (Done)

test/test_2D_scene.py
alexisyslau: WFOV w/ SPC in band 4 right? We have tested it with a disk model and prf cube
image

corgisim/scene.py
alexisyslau: This question has poked me to look deeper into `fftconvolve`. 

I started looking into the current logic of how we define the array size and how we are combining it on the detector. 

Under the current logic, there are 2 images:
host star image: defined by `output_dim`
convolution object: defined by either the input object size or the input prf size.

Array size matters, but the object and PRF do not need to have the same array dimensions.  SciPy’s `fftconvolve` accepts differently sized arrays and computes their full linear convolution internally. The selected mode then determines which part of the convolution is returned. 

We use `mode="same",` for which the output has the same dimensions as the first input. Previously, the object was the first input, so the output followed the object dimensions. Meaning if user uses a larger array size than the prf itself, the array size of the convolution would follow the object. 

`
(102, 102) obj * (51, 51) prf --> (102, 102) obj # I do not think we want that? 
`

The original implementation also forced every PRF slice to match the object dimensions before convolution. A smaller PRF was zero-padded, while a larger PRF was cropped. Neither operation is required by fftconvolve. Padding adds no physical PRF information, while cropping can remove meaningful PRF wings and should only be performed deliberately.

Looking closer into the size issue, I found this weird edge effects: 
`                conv += fftconvolve(weighted_scene, prfs_array[prf_idx], mode="same")
`
log scale
Image

normal scale
Image

I initially susptected the zero-padding caused the four semicircular structure near the image edges. Further testsing showed the issue is more subtle. 

The native 51 x 51 prf contains non zero signal at their boundaies, meaning the prf are already spatially truncated. When forcing the output of the convolution to have the same array size of the object, we will see  the outer regions beyond prf size that has the wrap structure outside of the fov. I think to completely eliminate this problem, we should generate a physically larger PRF cube at the same angular sampling, fourier shifting and convlution on that larger grid, confirm we do not have wrapping effects at the boundary, and crop it to the approriate size of the output image. 

To avoid accendicatlly having the bigger array than whats needed and matching the final output dimension of the host star image, the prf should be placed first. 
`                conv += fftconvolve(prfs_array[prf_idx], weighted_scene,mode="same")
`
For example, the current object is 102  x  102 , while each PRF slice is 51  x  51. We now retain the PRF at its native size and place it first in the convolution. SciPy uses the larger object when computing the convolution but returns the central  51  x  51 region, matching the PRF dimensions. The object is not cropped before convolution; only the final convolution result is cropped.

object: 

Image

conv (log scale)

Image

conv (normal scale)

Image

The important physical requirement is that the object and PRF share the same angular pixel sampling. The mode="same" option does not perform any resampling. Therefore, ensuring compatible spatial sampling remains the user’s responsibility, although I can confim the object array may be larger than the PRF slices (or smaller). 

Allowing a larger object also revealed a separate problem in the field-dependent PRF interpolation. This concerns the spatial positions represetned by the PRF slices, rather than the dimensions of each PRF slices. 

With` interpolate_prfs=True,` some object pixels can lie beyond the maximum radius covered by the PRF-position grid. In this case, the radial interpolation factor, alpha,  can exceed one. This causes terms containing \(1-\alpha\) to become negative, producing negative interpolation weights and consequently negative flux in parts of the output. I addressed this by constraining alpha to the valid interpolation range [0,1]`. 

Object pixels beyond the available PRF-position grid retain their flux but are assigned the outermost available radial PRF. This boundary treatment prevents non-physical extrapolation and negative interpolation weights. In the current example, the PRF grid appears to provide sufficient radial coverage for the region containing meaningful object flux, so this boundary condition should have little or no effect on the result. It remains an important safeguard for user-generated PRF cubes, where the radial coverage may not extend far enough to cover all non-zero scene emission. Ideally, the code should detect this condition and warn the user.

Finally, returning a PRF-sized disk image actually is another dimension requirement when the disk is combined with the host-star iamge.  We should check this before doing any sort of convolution: 

```
# PRF cube dimension (assumed to be the same as the output dimension)
prf_dim = fits.getdata(prf_path).shape[1]  

if output_dim != prf_dim:
    raise ValueError("Output dimension does not match PRF dimension")

```

This provides a clear error before convolution rather than an unexpected failure when the disk and host-star iamge are combined.

alexisyslau: My current guess is that the structures arise from finite-grid boundary effects, but this has not yet been confirmed. To test this, I can generate physically larger PRF stamps and object grids at the same angular sampling, perform the convolution on the larger grid, and crop the central 51*51 science region. Comparing its recovered flux with the current result should determine whether cropping excludes flux outside the intended field or whether the boundary effects also affect the photometry within it. For now, I think the current implementation is sufficient for generating scene data products.

maxwellmb: Let me try to summarize my understanding: 
1) If your image is too big, then you might not have sampled enough PRFs far enough out and things get a bit screwy. You can somewhat mitigate this by limited the value of some parameter alpha. I'm not sure I know what the alpha values are, so I don't fully understand what you're constraining there. 
2) The output dimensions are based on whatever you put first. 
3) If your PRFs are too small, then the convolution isn't perfect further out. This is in part because there are PRF wings. 

My thoughts: 
1) I think we _should_ in fact allow the final size to be determined by the science image. The choice of PRF size might depend on the fidelity to which you care about modeling those extended wings and a user might instead want to save time by generating smaller PRFs. In that case we wouldn't want the disk to be truncated. We could add some words in the tutorial about the tradeoffs. 
2) For any given mode, do we ensure (or gently force it so) that the PRFs are sampled from inside the coronagraph to outside the field mask?
3) Now that I think about it we actually should maybe think about forcing the output dimensions to be at least the dimensions of the field mask (i.e. the dark hole). We could do this by padding the disk if it's too small. I think we should do this what do you think? 
4) How are we placing the image on the detector, is it by the corner of the array or by the center of the array?

test/test_2D_scene.py
maxwellmb: Yes, WFOV SPC with band 4. Is there a test in corgisim?

maxwellmb: If not, maybe we make an issue about it.

maxwellmb (commented):
Couple last things. Mainly just resolving the data array size thing.

corgisim/scene.py
alexisyslau: > 1. If your image is too big, then you might not have sampled enough PRFs far enough out and things get a bit screwy. You can somewhat mitigate this by limited the value of some parameter alpha. I'm not sure I know what the alpha values are, so I don't fully understand what you're constraining there.

Yes, you are right about that. Here, alpha describes where a scene pixel lies radially between two sampled PRF locations. Normally, it ranges from 0 to 1 when the PRF-position grid adequately covers the fov. If the scene flux extends beyond the outermost sampled PRF position, alpha can exceed 1, producing negative interpolation weights and therefore negative flux through `weighted_scene = obj * weight_map`. Because I encountered this case during testing and traced the negative flux back to alpha, I now constrain alpha from 0 to 1 . Pixels beyond the grid therefore use the outermost available radial PRF. This prevents non-physical extrapolation, although generating PRFs farther out would be the more physically accurate solution.

maxwellmb: Maybe we trigger a warning if that happens?

alexisyslau: > 2. The output dimensions are based on whatever you put first.
> 3. If your PRFs are too small, then the convolution isn't perfect further out. This is in part because there are PRF wings.

Both are correct. With `mode="same"`, the dimensions of the convolution output follow those of the first input. Smaller PRFs truncate more of the extended PRF wings, which can underestimate flux and reduce accuracy, particularly near image boundaries.
> 
> My thoughts:
> 
> 1. I think we _should_ in fact allow the final size to be determined by the science image. The choice of PRF size might depend on the fidelity to which you care about modeling those extended wings and a user might instead want to save time by generating smaller PRFs. In that case we wouldn't want the disk to be truncated. We could add some words in the tutorial about the tradeoffs.

Ok, understood. This makes sense. I changed the convolution to use the science image as the output size reference without resizing the PRFs. Users can therefore choose smaller PRFs to reduce generation time, memory use, and convolution cost, at the expense of truncating more of the extended PRF wings.
I made the following changes:
- Added an explanation of PRF dimensions and output size to the optics-setup section of the disk tutorial.
- Swapped the argument order in `_convolve_with_prfs()` so that the science image determines the output dimensions.
- Updated the function’s Notes section to state that the output dimensions follow those of `obj`.

> 2. For any given mode, do we ensure (or gently force it so) that the PRFs are sampled from inside the coronagraph to outside the field mask?

Not currently. The PRFs are generated using the selected optics keywords, but they're radial. and azimuthal sampling locations are still defined by the user. `build_radial_grid()` accepts users-provided IWA, OWA, step sizes and max radius, and these values are not currently validated against the selection mode's boundaries. 

I think we should provide a mode-dependent recommended sampling range and issue warnings when the PRF grid does not contain samples within the IWA or does not extend beyond the required outer boundary. I would prefer warning rather than errors because user might internationally generate a restricted PRF grid for faster calculation or for applications other than extended-scene convolution . 

To implement this, we need IWA and OWA for each modes we are supporting. Ideally, these should be stored in the something like `constant.py`. 

> 3. Now that I think about it we actually should maybe think about forcing the output dimensions to be at least the dimensions of the field mask (i.e. the dark hole). We could do this by padding the disk if it's too small. I think we should do this what do you think?

Yup. agreed. We can define a minimum science image dimension for each mode, for example, 51*51 for HLC Band 1 NFOV. If the input disk is smaller, centre pad with zeros to the minimum dimensions before convolution.  If it is larger, it will remain unchanged. 

This preserves the behaviour discussed above: the science image determines the output size,  subject to the mode minimum.that ensures the full dark-hole region is represented. I think these minimum dimensions should be stored alongside the IWA and OWA values in `constants.py`. 

I have added a warning in the tutorial when the input disk is smaller than the mode minimum. Automatic padding is not implemented yet, should be added to `simulate_2d_scene() `before normalisation and convolution. I propose to implement both 2. and 3. togethe in a follow-up PR. 


> 4. How are we placing the image on the detector, is it by the corner of the array or by the center of the array?

When a full detector frame is requested, the convolution array is placed by its center, not by a corner. Its centre pixel is aligned with `(loc_x, loc_y),` which defaults to detector pixel (512, 512). This behaviour is implemented by generate_detector_image() and place_scene_on_detector(). For a 51 × 51 array placed at (512, 512), its lower corner is placed at (487, 487), and its centre pixel is placed at (512, 512).

alexisyslau: The science image dimensions are used when setting `output_dim`: a disk larger than the mode’s dimensions increases the output size, while a smaller disk leaves the default unchanged. Automatic center-padding is not yet implemented. Is this what we wanted to have for setting the output_dim?

maxwellmb (approved):

PR #71, opened 2025-06-23
add roll angle in simulation

zjwhale:
## Purpose
Add support for telescope roll angle in the `CorgiOptics` class.

## Summary of Changes
**Major updates** in `instrument.py` and `outputs.py`:

1. **Initialization**:  
   Added a `roll_angle` keyword argument to the `CorgiOptics` initializer. This value is stored as a global variable `self.roll_angle`.

2. **Point Source Injection**:  
   In the `inject_point_source` function, companion positions are now rotated relative to the frame center (i.e., the center of the coronagraph) according to `self.roll_angle`.

3. **FITS Header**:  
   The roll angle is written to the primary FITS header as `prihdr['ROLL']` in the Level 1 data product.

4. **Tutorial**:  
   Updated the tutorial notebook to demonstrate the effect of `roll_angle` and how to use it in simulations.


zjwhale:
I see comments about Roll angle from Vanessa and Jason in sim meeting on Jun 26 2025. Move them here for record. 

Vanessa:
"not totally sure I understood the question about roll centers, but in case it’s useful: from an operational standpoint, a roll is just a type of telescope slew, so the star would be re-acquired behind the coronagraph mask. ie: the star would move to the mask, rather than the mask needing to move to the star."

Jason:
"Also: I don't believe we will be rotating for any of the non-coronagraphic calibration images. we will dither, but not roll."

test/test_L1_product_fits_format.py
maxwellmb, line 1: You also need a test to show that the point source you injected is actually injected at the correct roll angle.

corgisim/instrument.py
```diff
@@ -373,10 +377,20 @@ def inject_point_sources(self, input_scene, sim_scene=None, on_the_fly=False):
...
                 proper_keywords_comp = self.proper_keywords.copy()
+                if self.roll_angle == 0.0:
+                    x_offset_source_j = point_source_x[j]
+                    y_offset_source_j = point_source_y[j]
+                else:
+                    ##### rotate the telescope based on Roll angle
+                    sep_source_j = np.sqrt(point_source_x[j]**2 +  point_source_y[j]**2)
+                    PA_source_j = calculate_PA(point_source_x[j], point_source_y[j]) ## rad
+                    x_offset_source_j = sep_source_j * np.sin( PA_source_j + np.deg2rad(self.roll_angle) )
+                    y_offset_source_j = sep_source_j * np.cos( PA_source_j + np.deg2rad(self.roll_angle) )
```
maxwellmb, line 388: This is fine, but I think you could have also done this with a rotation matrix (the signs might be messed up depending on which way things are rolling): 
x_new = x_old*cos(roll) + y_old*sin(roll)
y_new = -x_old*sin(roll) + y_old*cos(roll)

corgisim/instrument.py
```diff
@@ -373,10 +377,20 @@ def inject_point_sources(self, input_scene, sim_scene=None, on_the_fly=False):
             for j in range(len(point_source_spectra )):
             
                 proper_keywords_comp = self.proper_keywords.copy()
+                if self.roll_angle == 0.0:
+                    x_offset_source_j = point_source_x[j]
+                    y_offset_source_j = point_source_y[j]
+                else:
+                    ##### rotate the telescope based on Roll angle
```
maxwellmb, line 384: have you confirmed with Alex (or someone else) the direction that the roll angle keyword is defined relative to the focal plane (i.e. does positive mean clockwise or counter-clockwise? 

Also, out of curiosity, have you confirmed whether we expect to get images with East being counter-counterclockwise of North, or the opposite? We should figure that out now - its relevant to the rotation.

maxwellmb (commented):
I think we need to add to the documentation and tutorial that point source x and y coordinates are now delta_ra and delta_dec.

test/test_L1_product_fits_format.py
zjwhale: Just added it

corgisim/instrument.py
```diff
@@ -153,6 +156,19 @@ def __init__(self, cgi_mode = None, bandpass= None,  diam = 236.3114, optics_key
...
+                # Rotate slit offsets to match the companion after telescope roll.
+                # Convention: x=East, y=North. Telescope roll = +θ.
+                # Because we rotate the *companion coords*, use the opposite sense: θ_comp = -θ_tel.
+                rotate_angle = -1 * self.roll_angle
+                sep_slit = np.sqrt(self.slit_x_offset_mas**2 +   self.slit_y_offset_mas**2)
+                PA_slit = calculate_PA(self.slit_x_offset_mas, self.slit_y_offset_mas) ## rad
+                x_offset_slit= sep_slit * np.sin( PA_slit + np.deg2rad(rotate_angle) )
+                y_offset_slit = sep_slit * np.cos( PA_slit + np.deg2rad(rotate_angle) )
+                self.slit_x_offset_mas, self.slit_y_offset_mas = x_offset_slit,y_offset_slit 
+
```
maxwellmb, line 171: I think if you want to include this you should add a spec mode keyword that allow the user to toggle this automatically-adjusting behavior on and off (and by default have it off). In general I don't like the code to automatically adjust things for people by default, but its a fine feature to have. 

@neilzim if a user did want it automatically adjusted, is this the way to do it?

corgisim/instrument.py
```diff
@@ -553,13 +572,26 @@ def inject_point_sources(self, input_scene, sim_scene=None, on_the_fly=False):
...
+                    y_offset_source_j = point_source_y[j]
+                else:
+                    # Rotate the telescope based on roll angle
+                    # Convention: x=East, y=North. Telescope roll = +theta.
+                    # Because we rotate the *companion coords*, use the opposite sense: θ_comp = -θ_tel.
+                    rotate_angle = -1 * self.roll_angle
+                    sep_source_j = np.sqrt(point_source_x[j]**2 +  point_source_y[j]**2)
+                    PA_source_j = calculate_PA(point_source_x[j], point_source_y[j]) ## rad
+                    x_offset_source_j = sep_source_j * np.sin( PA_source_j + np.deg2rad(rotate_angle) )
+                    y_offset_source_j = sep_source_j * np.cos( PA_source_j + np.deg2rad(rotate_angle) )
```
maxwellmb, line 587: I'm going to let this through for now, but it looks like there's a lot of re-used code between excam and spec that we should pull out of the "if statement". Can you make an issue about it to clean it up?

test/test_roll_angle.py
```diff
@@ -0,0 +1,120 @@
...
+import matplotlib.pyplot as plt
+import numpy as np
+import proper
+import roman_preflight_proper
+import pytest
+import cgisim
+
+
+
+def test_roll_imaging():
```
maxwellmb, line 13: Can you please add some docstrings to describe what the tests setup and what they actually test?

corgisim/instrument.py
```diff
@@ -963,3 +1012,40 @@ def define_EMCCD(self, emccd_keywords=None ):
         return emccd
 
 
+def calculate_PA(dx, dy):
```
maxwellmb, line 1015: it's just ocurred to me that perhaps a better way to do this would be to have a source have [dx, dy, sep and PA] properties and changing any one of them would update the others appropriate (via a setter class). Let's add an Issue to make that a future upgrade, but let this go through now.

maxwellmb (changes requested):
A few minor comments and a couple issues for future upgrades

PR #87, opened 2025-08-12
Update python-app.yml

helloWorlds0:
Add durations to see the slowest tests


zjwhale:
Should we  separate the tests into two categories?
1. small and fast test units that run every time when people push PR
2. large end-to-end tests that run on weekend

zjwhale:
I found  setting `oversampling_factor = 1` could  reduce the running time. The default is `oversampling_factor = 7`.

PR #89, opened 2025-08-26
Add polarimetry mode

everlastingEric:
# Overview
Add the ability to simulate polarimetric observations in Corgisim

# List of changes
- Added `'POL0'` and `'POL45'` as supported prism values in `optics_keywords`, which correspond to the 0 and 45 degree Wollaston prisms found on the DPAM. These two prism parameters are used in excam coronagraph mode to generate polarimetric images. 
- Updated `get_host_star_psf()` and `inject_point_sources()` functions in the `CorgiOptics`  class to generate 0/90 degree or 45/135 degree polarized image pairs of the speckle field and point sources. The image pairs are stored as 2 by image_size_x by image_size_y arrays inside the scene object. 
- Updated the `CorgiDetector` class with the ability to place the polarized images appropriately onto the detector frame, based on what polarization basis is specified on the scene header. 
- Added new supported `polaxis` parameters in addition to existing ones defined as follows:
    - -10: Incoherent average of `polaxis` = -2, -1, 1, 2; output is an unpolarized image with the full polarization aberration model 
    - -4: -45 deg in, -45 deg out
    - -3: -45 deg in, +45 deg out
    - 3: +45 deg in, +45 deg out
    - 4:+45 deg in, -45 deg out
- Added new function `generate_full_aberration_psf` in the `CorgiOptics` class to compute polaxis = -10
- Updated `outputs.py` to write the appropriate DPAM values to the output L1 image header depending on which wollaston is used
- Added a tutorial notebook called `polarimetry_demo.ipynb` in the example folder as a guide on simulating polarimetric observations
- Updated `README.md`, `python-app.yml`, and `.readthedocs.yaml` with new instructions to install from the modified cgisim repository containing new polaxis parameters instead of the default one on sourceforge. 
- Created a new test file `test_polarimetry.py` to verify the functionality of these changes. However it currently takes 10+ minutes to run, which might be an issue that needs to be addressed if we're trying to shorten the time it takes for the automatic testing to run.


maxwellmb (approved):
Great stuff Eric!

PR #96, opened 2025-08-29
add a point source to cpgs scenario

helloWorlds0:


maxwellmb (changes requested):
It looks like the docstring for `generate_observation_scenario_from_cpgs` was maybe out of date and includes loc_x and loc_y instead of point_source_info. 

Otherwise this looks good

PR #114, opened 2025-09-28
Add_roll_angle_new

zjwhale:
## Purpose
Add support for telescope roll angle handling in the `CorgiOptics` class.

## Summary of Changes
**Major updates** to `scene.py`, `instrument.py`, and `outputs.py`:

1. **Coordinate system changes**  
   - Updated companion position inputs in `Scene.py` from `(dx, dy)` to `(dRA, dDEC)`.  
   - Similarly, slit position inputs for spectroscopic mode are now defined in `(dRA, dDEC)` instead of `(dx, dy)`.

2. **Initialization**  
   - Added a new keyword argument `roll_angle` to the `CorgiOptics` initializer.  
   - Stored internally as `self.roll_angle`.  
   - **Definition**:  
     `roll_angle : float, optional`  
     Telescope roll angle in degrees (0–360). Defined as the rotation of the excam coordinate system (X, Y) relative to sky coordinates (RA, DEC). Positive values rotate counter-clockwise. Default is 0°, corresponding to the standard orientation (North up, East left).

3. **Point source and slit position handling**  
   - In `inject_point_source`, companion positions are converted from sky coordinates `(dRA, dDEC)` to excam coordinates `(dx, dy)` using `self.roll_angle`.  
   - Slit positions are also converted using the same transformation.

4. **FITS header update**  
   - The telescope roll angle is now written into the primary FITS header of the Level 1 data product as:  
     ```python
     prihdr['ROLL'] = self.roll_angle
     ```

5. **Tutorial**  
   - Extended the tutorial notebook to demonstrate the effect of `roll_angle` and show how to configure it in simulations.

6. **Tests**  
   - Added `test_roll_angle.py` to validate coordinate transformations and FITS header output.

### DIFF
--- a/corgisim/instrument.py
+++ b/corgisim/instrument.py
@@ -140,6 +140,8 @@ class CorgiOptics():
                 'slit': 'None', # named FSAM slit
                 'slit_ra_offset_mas': 0.0, # offset of slit position from star on sky coordinate, in mas
                 'slit_dec_offset_mas': 0.0, # offset of slit position from star on sky coordinate, in mas
+                'slit_x_offset_mas': None, # offset of slit position from star on excam coordinate, in mas
+                'slit_y_offset_mas': None, # offset of slit position from star on excam coordinate, in mas
                 'prism': 'None', # named DPAM prism
                 'wav_step_um': 1E-3, # wavelength step size of the prism dispersion model, in microns 
                 'fsm_x_offset_mas': 0.0, # FSM x offset in mas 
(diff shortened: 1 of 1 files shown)


test/test_roll_angle.py
```diff
@@ -0,0 +1,130 @@
...
+import numpy as np
+import proper
+import roman_preflight_proper
+import pytest
+import cgisim
+
+
+
+def test_roll_imaging():
+    Vmag = 8
```
maxwellmb: please add some details about what this tests and how it does it

test/test_roll_angle.py
```diff
@@ -0,0 +1,130 @@
...
+    sep2_slit = np.sqrt(x2_slit**2+y2_slit**2)
+    PA2_slit =  np.rad2deg(np.arctan2(y2_slit, x2_slit))
+
+    dPA = PA1-PA2
+    assert sep1 == pytest.approx(sep2,abs=0.1), (f"Separation check failed: expected  {sep1:.3f} mas, got {sep2:.3f} mas).")
+    assert dPA == pytest.approx(roll_angle,abs=0.1), (f"Roll-angle check failed: expected  {roll_angle:.3f} degree, got roll={dPA:.3f}degree).")
+
+    dPA_slit = PA1_slit-PA2_slit
+    assert sep1_slit == pytest.approx(sep2_slit,abs=0.1), (f"Separation for slit check failed: expected  {sep1_slit:.3f} mas, got {sep2_slit:.3f} mas).")
+    assert dPA_slit == pytest.approx(roll_angle,abs=0.1), (f"Roll-angle for slit check failed: expected  {roll_angle:.3f} degree, got roll={dPA_slit:.3f}degree).")
```
maxwellmb: I think the relevance of this test will depend on @neilzim's response above.

maxwellmb (changes requested):
A few comments here. Thanks for your patience

PR #119, opened 2025-09-30
add option to save observing sequence as fits

helloWorlds0:


corgisim/observation.py
```diff
@@ -48,10 +48,22 @@ def generate_observation_sequence(scene, optics, detector, exp_time, n_frames, f
             sim_image = detector.generate_detector_image(sim_scene,exp_time)
             simulatedImage_list.append(sim_image)
     else:
+        if save_as_fits:
+            # Save the images as fits in output_dir if specified, in corgisim/test/testdata if not
+            # Simulation needs to be full frame to be written as L1
+            if output_dir == None:
+                local_path = corgisim.lib_dir
+                outdir = os.path.join(local_path.split('corgisim')[0], 'corgisim/test/testdata')
```
maxwellmb, line 56: if a user doesn't provide an output directory I suggest either, we throw an error if no directory is given OR we print something out that tells people where it's being saved, just so that its not hidden

zjwhale (approved):

PR #126, opened 2025-11-03
101 website should have the tutorial notebooks

helloWorlds0:
The documentation website ([here for the version on this branch](https://corgisim--126.org.readthedocs.build/en/126/)) now has the notebooks.


helloWorlds0:
> Could we put the " corgisim" in the last place and called it as "detailed corgisim API"? Some example can be found for a package called spaceKLIP here: https://spaceklip.readthedocs.io/en/latest/spaceKLIP.html

Done! It's not as polished as spaceKLIP but we have the tutorials first, then the API, with headings for each

maxwellmb (changes requested):
This is awesome! Here are some thoughts: 

- The installation page says to install cgisim from sourceforge, but I think it should point to our own repo. 
- The installation page also has the "Run the first simulation" tacked on to it, but do we want it as a separate page, like the "demo of the polarimetry mode" page? 

There were a couple other comments from the meeting today. The things I remember:
- Move the website to the top of the readme
- Add a note to the contributor guide about how to add a new tutorial page to the website.

helloWorlds0:
I took the remarks into account. I also clear the outputs from the notebook, which means that they are now run each time the documentation is build. The goal is o know when a change in the code necessitates a change in the tutorials. The downside is that the documentation takes much longer to build.   

>     * The installation page also has the "Run the first simulation" tacked on to it, but do we want it as a separate page, like the "demo of the polarimetry mode" page?

I think we should leave it as-is. The menu on the left still lists them both separately (see image) but it's a reminder for newcomer to do a proper install before running the first simulation. 
image

PR #141, opened 2025-12-11
Add parameter tracking to grid building functions and PRF header generation

alexisyslau:
Enhanced grid building functions to automatically capture and return input parameters, and added a new function to generate FITS headers for the PRF cubes with complete metadata (?). 

**Changes**
1. Both `build_radial_grid()` and `build_azimuth_grid()` will return `params` along with the current output. 

- `params` is a dictionary that contains all input parameters used to generate the grid --> tracking without requiring the users to manually record values 

2. Added `_generate_prf_header()`function, which stores the following information:
- Radial grid parameters
- Azimuthal grid parameters 
- optic keywords (i.e. which imaging mode, bandpass e.t.c) 
- dm solution information is stored separately because dm keys in the optics_keywords are fits file -> storing `rootname` instead, just to keep track of which dm solution we are using 

Issues: 
1. Need to confirm which unit we are using in the PRF cube. Currently marked as '???' 

Note:
This PR is based on #64. It modifies the return signatures of build_radial_grid() and build_azimuth_grid() to return grid, params instead of just grid. This PR needs to be merged in `2D_convolution` before merging #64

### DIFF
--- a/corgisim/convolution.py
+++ b/corgisim/convolution.py
@@ -15,67 +15,32 @@ PIXEL_SCALE_ARCSEC = 0.0218   # arcsec/pixel
 
 from astropy.io import fits
 
-def _generate_prf_header(radial_param, azimuthal_param, optics_keywords, dm_solution) -> fits.Header:
+def _generate_prf_dictionary(radial_param, azimuthal_param, dm_solution) -> dict:
     """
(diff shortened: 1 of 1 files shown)


alexisyslau:
It's easier to keep track of all the changes if we stick to the original PR. will move the changes over

PR #164, opened 2026-01-08
Add time series based on os11

AlexBidot:
Enhancement to add time variable speckles based on the wavefront error expectation of os11 scenario.
The generated sequence will be divided the same as os 11 with alternating the roll angle and observation cycle between reference and target star. Outputs will be saved both with and without the emccd noise.

**Main changes:**
Files created: 
wavefront_estimation.py : script to add the function needed to retrieve and interpolate the os11 wavefront error. Also the main function to generate an observation sequence without CPGS is in there (generate_time_series_from_os11)
os11_example.py : main to run the generate_time_series_from_os11 as an example.

Files changed:
observation.py : a few lines were modified or added to support wavefront error with CPGS sequence observation.
scene.py : a few lines were added to support zwfs and make sure we have the good optics_keywords.

Limitations:
No sequence can be longer than the os11 for interpolation reasons. The wavefront errors are not yet scaled with the slew but work is in progress to add more flexibility.


corgisim/wavefront_estimation.py
```diff
@@ -0,0 +1,347 @@
...
+    """
+    noiseless_image = noiseless_image_hdu.data
+    saturation = 12000
+    threshold = 0.8 * saturation
+    max = np.nanmax(noiseless_image)
+    emgain = threshold/max
+
+    return emgain
+
+def generate_time_series_from_os11(ref_star_properties, target_star_properties, point_source_info, frame_exp, total_exp_time, outdir, cor_type='hlc', bandpass = '1F', polaxis=10, dm_case='3e-8'):
```
zjwhale: `generate_time_series_from_os11()` feels like it belongs in `observation.py` rather than `wavefront_estimation.py`.

The other functions in this module (`get_reference_aberration()`, `get_science_acquisition()`, `get_drift()`, and `get_jitter()`) read as lower-level OS11/WFE utility helpers that can be reused by `instrument.py` or `observation.py`. In contrast, `generate_time_series_from_os11()` is a higher-level end-to-end workflow: it builds scenes, configures optics/detector objects, loops over cycles/rolls/frames, generates images, and writes outputs. That seems more consistent with the responsibilities of the observation layer.

zjwhale (changes requested):
Sorry it took me a while to review this. The PR looks good to me overall. My main comment is that it may be better to separate the ZWFS simulation work from the time-series simulation work into different PRs, so this PR can stay focused on the time-series functionality.

Another comments is that we probably want to have the simulating time-series function with currently existing function in `observation.py`, such as `generate_observation_sequence()` or `generate_observation_scenario_from_cpgs()` instead of creating a new function `generate_time_series_from_os11()`

corgisim/observation.py
```diff
@@ -2,9 +2,13 @@
...
 import os
-from corgisim import scene, instrument, inputs, observation, outputs
+from corgisim import scene, instrument, inputs, observation, outputs, wavefront_estimation
+from corgisim.wavefront_estimation import get_drift
 
-def generate_observation_sequence(scene, optics, detector, exp_time, n_frames, save_as_fits= False, output_dir=None, full_frame= False, loc_x=None, loc_y=None):
+from astropy.io import fits
+
+
+def generate_observation_sequence(scene, optics, detector, exp_time, n_frames, save_as_fits= False, output_dir=None, full_frame= False, loc_x=None, loc_y=None, zernike_index=None, zernike_value_m=None):
```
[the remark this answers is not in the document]
maxwellmb: In what context are these new keywords used?

maxwellmb (changes requested):

corgisim/observation.py
```diff
@@ -41,13 +45,19 @@ def generate_observation_sequence(scene, optics, detector, exp_time, n_frames, s
...
     simulatedImage_list = []
-    
+
     if full_frame == False :
         for i in range(0, n_frames):
-            sim_image = detector.generate_detector_image(sim_scene,exp_time)
-            simulatedImage_list.append(copy.deepcopy(sim_image))
+            if zernike_index is not None:
+                optics.optics_keywords.update({'zindex': zernike_index, 'zval_m': zernike_value_m[i]})
+            sim_scene = optics.get_host_star_psf(scene)
```
helloWorlds0, line 55: This simulation has already been done at the beginning of the function

PR #177, opened 2026-02-07
change the roman preflight lib

zjwhale:
In minimal_test.yml, we were previously downloading Roman preflight v2.0.0, which now causes issues because John has updated the package to v2.0.1. Besides, we’re using a fork that includes the polaxis parameters (needed to model polarization aberrations on the ±45° output axes), I’ve switched the YAML to point to our forked repo instead.

That said, we still need to sync our fork with John’s v2.0.1 updates—with the new DM files.

### DIFF
--- a/.github/workflows/minimal_test.yml
+++ b/.github/workflows/minimal_test.yml
@@ -38,8 +38,8 @@ jobs:
         cd ..        
     - name: Download and install Roman preflight
       run: |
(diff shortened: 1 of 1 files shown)


maxwellmb:
It looks like the change is to install cgisim, but before it was just the proper module. Is this ok? (The tests seem to pass, but I wanted to check)

maxwellmb (approved):
Ah, I see minimal_test.yml still has the proper stuff in there anyway.

PR #181, opened 2026-02-10
Update README.md

helloWorlds0:
Making it a bit clearer, several people had an issue because they didn't install cgisim_cpp during the workshop


helloWorlds0:
> This is in draft form. Ready to go?

It should be resolved by #183 . keeping it in draft form until that one is merged, then I'll delete it

Issue #167, opened 2026-01-09
Tutorial for finite stellar diameter and jitter


gershrange:
Need to write a tutorial for how to use the finite stellar diameter and jitter model.

PR #188, opened 2026-02-16
Create finite_stellar_diam_tutorial.ipynb

gershrange:
Adding a notebook with a tutorial explaining how to represent the host star as a finite-diameter disk instead of a point source. This pull request partially addresses issue #167 Tutorial for finite stellar diameter and jitter.


maxwellmb (changes requested):
Only small comment: 
- Do you want to make a comment somewhere about runtime and how one might expect it to change?

maxwellmb (approved):

Issue #144, opened 2025-12-11
update stellar spectrum and companion spectrum


zjwhale:
We currently has stellar spectrum as black body function and companion spectrum as flat line. 

Need to updated the two with more realistic spectrum

PR #203, opened 2026-04-01
scene accepts user-provided planet spectrum

neilzim:

- Changes to get_off_axis_source_spectrum() function in scene.py to enable custom spectrum to address Issue #144 
- Spectroscopy demo notebook (examples/spec_slit_prism_demo.ipynb) updated to show usage
- Added unit tests for the new feature in test/test_off_axis_source_spectrum.py

- [X] I have verified that test_minimal.py passes and I have added a test there if appropriate.

## Checklist before merging


corgisim/scene.py
```diff
@@ -448,10 +453,14 @@ def get_off_axis_source_spectrum(self, vmag, magtype, spectrum=None):
         vmag : list of float
             Desired V-band magnitudes for the off-axis sources.
         magtype : list of str
-            Magnitude system used for each source. 
+            Magnitude system used for each source.
         spectrum : list or None, optional
             Custom input spectra. If None, flat spectra will be used and scaled to match `vmag`.
             Custom spectra are not yet supported and will raise an error if provided.
```
helloWorlds0: We can delete this line! :)

neilzim:
@zjwhale @helloWorlds0 Following up on the additional test you requested during yesterday's sims telecon: we could add a test to test_spec.py where we run the proper model to make prism images of the star (unocculted) and planet. We can verify the band-integrated flux ratio in the prism images is consistent with the input planet and star source models. Would this suffice? If you have something else in mind, let me know.

helloWorlds0:
> @zjwhale @helloWorlds0 Following up on the additional test you requested during yesterday's sims telecon: we could add a test to test_spec.py where we run the proper model to make prism images of the star (unocculted) and planet. We can verify the band-integrated flux ratio in the prism images is consistent with the input planet and star source models. Would this suffice? If you have something else in mind, let me know.

@neilzim  I think that would be good! 
I would also like a test in test_minimal (even if it’s only changing the test_excam_mode to include a custom spectrum on the companion). In the same spirit, test_off_axis_source could be modified/extended to have a custom spectrum, wit the resulting image being compared to the cgisim simulation. @zjwhale what do you think?

neilzim:
The tests you requested are ready to be reviewed.

helloWorlds0:
> The tests you requested are ready to be reviewed.

Thanks Neil, I'm a bit busy with the CAR right now but I'll review it by the end of the week or beginning of the next at the latest

test/test_off_axis_source.py
```diff
@@ -96,6 +106,7 @@ def test_off_axis_source():
     ## the test is for noise-free simulation
     assert  image_star_corgi  == pytest.approx(image_star_cgi, rel=0.5)
     assert  image_comp_corgi  == pytest.approx(image_comp_cgi, rel=0.5)
+    assert  customspec_image_comp_corgi  == pytest.approx(image_comp_cgi, rel=0.5)
```
helloWorlds0: This relative tolerance is rather high (although I recognize that it's the same as the other images; they should also be lower). A relative tolerance of 0.15 seems more appropriate to me, what do you think @zjwhale ?

helloWorlds0:
looks good to me, just a docstring to update in get_off_axis_source_spectrum and the relative tolerance in test_off_axis_source. @zjwhale let me know if you have other remarks

helloWorlds0 (approved):

Issue #205, opened 2026-04-09
Replace ROLL in header with PA_APER


gershrange:
In the DRP, ROLL is being deprecated and replaced by PA_APER. Line 1444 in instrument.py needs to be modified to accommodate this change.

Currently, the data generated by corgisim places the roll angle in ROLL and leaves PA_APER at 0. When this data is passed through the DRP, the final L4 image shows multiple images of the companion(s).

gershrange:
Closed by PR #206

PR #206, opened 2026-04-09
Header update: Replace ROLL with PA_APER

gershrange:
For consistency with the DRP, this PR replaces 'ROLL' in the prihdr with 'PA_APER'. Without this change, the DRP does not recognize roll angles correctly, so datasets containing multiple rolls lead to an L4 image with multiple copies of the companion(s).
This PR addresses Issue #205. 

- [X ] I have verified that test_minimal.py passes and I have added a test there if appropriate.

## Checklist before merging

### DIFF
--- a/test/test_L1_product_fits_format.py
+++ b/test/test_L1_product_fits_format.py
@@ -167,7 +167,7 @@ def test_L1_product_fits_format():
     os.remove(f)
 
     ####################################################################################################
-    #### testing the non-defalut(input) value pass to header
+    #### testing the non-default(input) value pass to header
     Vmag = 8
     sptype = 'G0V'
     cgi_mode = 'excam'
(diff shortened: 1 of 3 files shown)


helloWorlds0 (changes requested):
the tests (test/test_L1_product_fits_format.py at least) also needs to be updated (check that 'ROLL' does not change but that PA_APER does)

gershrange:
test/test_L1_product_fits_format.py has been updated

test/test_L1_product_fits_format.py
helloWorlds0, line 93: The PA_APER should also be checked here (we can leave the ROLL keyword or delete it)

helloWorlds0: Also change it here

helloWorlds0 (commented):
@gershrange There are some other places where the keyword checks need to be changed

gershrange:
@helloWorlds0 I've added the keyword checks in the other locations. 
I'm leaving the checks that prihr['ROLL'] == 0.0 for now so that they can be used to verify that we have transitioned from using 'ROLL' to using 'PA_APER'. These checks will need to be removed if we stop including 'ROLL' in the headers, which might be worth doing to prevent confusion.

helloWorlds0 (approved):

PR #213, opened 2026-04-23
Realistic MJDSRT and MJDEND

juliamilton:
Updates outputs.py so that MJDSRT is set to the Julian date equivalent of the value that is in FTIMEUTC. MJDEND = MJDSRT + (Julian date equivalent of exposure time).

This was done in response to corgidrp issue 853: https://github.com/roman-corgi/corgidrp/issues/853 but was determined corgisim would be a better place for this fix.

## Checklist before merging

### DIFF
--- a/corgisim/outputs.py
+++ b/corgisim/outputs.py
@@ -67,10 +67,6 @@ def create_hdu_list(data, header_info, sim_info=None):
     exthdr['NAXIS1'] = data.shape[0]
     exthdr['NAXIS2'] = data.shape[1]
     exthdr['EXPTIME'] = header_info['EXPTIME']
-    # Calculate MJDSRT and MJDEND from FTIMEUTC timestamp and EXPTIME
-    mjd_start = Time(exthdr['FTIMEUTC']).mjd
-    exthdr['MJDSRT'] = mjd_start
-    exthdr['MJDEND'] = mjd_start + exthdr['EXPTIME'] / 86400.0
     exthdr['EMGAIN_C'] = header_info['EMGAIN_C']
     exthdr['EMGAIN_A'] = header_info['EMGAIN_C']  
     exthdr['KGAINPAR'] =  header_info['KGAINPAR']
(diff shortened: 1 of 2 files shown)


corgisim/outputs.py
```diff
@@ -66,6 +67,10 @@ def create_hdu_list(data, header_info, sim_info=None):
     exthdr['NAXIS1'] = data.shape[0]
     exthdr['NAXIS2'] = data.shape[1]
     exthdr['EXPTIME'] = header_info['EXPTIME']
+    # Calculate MJDSRT and MJDEND from FTIMEUTC timestamp and EXPTIME
+    mjd_start = Time(exthdr['FTIMEUTC']).mjd
+    exthdr['MJDSRT'] = mjd_start
+    exthdr['MJDEND'] = mjd_start + exthdr['EXPTIME'] / 86400.0
```
zjwhale: Looks good! Could we add a small test for the newly added headers to make sure they are written in the correct format? One simple option would be to extend the component test for overriding headers in `test_L1_product_fits_format()` in `test_L1_product_fits_format.py` : override `FTIMEUTC` in `overwrite_ext_keywords` and then check that the corresponding MJDSRT is set correctly.

   ```
 ### Test overwrite_pri_hdr and overwrite_ext_hdr with non-default values
    outputs.save_hdu_to_fits(sim_scene.image_on_detector,outdir=outdir, write_as_L1=True, 
                             overwrite_pri_keywords={'TARGET':'HD 141569A', }, 
                             overwrite_ext_keywords={'OPMODE': "Disco","FTIMEUTC":2025-01-01T00:00:00},
                             )
    assert exthdr['MJDSRT'] == 60676.0, f"Expected data 'MJDSRT' == 60676.0, but got {exthdr['MJDSRT'] }"
```

zjwhale (changes requested):

corgisim/outputs.py
juliamilton: thanks! i added the test and moved the code block to after the overwrite in case someone overwrites the FTIMEUTC value (like in the test!)

it passes locally

zjwhale (approved):

PR #217, opened 2026-04-29
disable the simultaneous use of focal plane mask with certain nd filters

zjwhale:
There are three ND filter options:

self.nd = 1: ND 2.25 @ FPAM
self.nd = 2: ND 4.75 @ FPAM
self.nd = 3: ND 4.75 @ FSAM

This PR disables the simultaneous use of ND filters 1 or 2 with the focal plane mask, since they occupy the same position in the optical path.

It also disables the simultaneous use of ND filter 3 with the Lyot stop.

### DIFF
--- a/corgisim/instrument.py
+++ b/corgisim/instrument.py
@@ -269,8 +269,6 @@ class CorgiOptics():
             if (optics_keywords_internal['use_fpm']==1) or (optics_keywords_internal['use_lyot_stop']==1) or (optics_keywords_internal['use_field_stop']==1):
                  warnings.warn('Warning: the pupil lens is inserted while one or more of the focal mask, Lyot stop, or field stop are also in use.', UserWarning)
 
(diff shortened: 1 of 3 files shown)


maxwellmb (changes requested):
Let's add a quick test somewhere to make sure this triggers properly.

zjwhale:
> t somewhere to make sure this trigg

I have added the test

maxwellmb:
is the use of opts_keywords{"nd"} documented anywhere?

zjwhale:
> is the use of opts_keywords{"nd"} documented anywhere?

I have just pushed a commit to add the instruction of neutral density filter in the tutorial.

maxwellmb (approved):
I think we need to add some documentation somewhere (perhaps in the docstrings) to describe what the .nd options are.

PR #220, opened 2026-04-30
add function to allow empty scene

zjwhale:
Title

Allow truly empty scenes to propagate through optics and observation

Description

This PR adds support for a truly empty scene, defined as a scene with no host star, no companions, and no 2D background.

Previously, scene.Scene() could be constructed, but the simulation pipeline still assumed a host star existed and failed once get_host_star_psf() tried to build a stellar observation from input_scene.stellar_spectrum.

This change makes empty scenes flow end-to-end by treating them as valid zero-flux inputs.


maxwellmb (changes requested):
Please add this functionality to an existing tutorial.

zjwhale:
> Please add this functionality to an existing tutorial.

I have updated the tutorial.

corgisim/observation.py
```diff
@@ -39,7 +39,7 @@ def generate_observation_sequence(scene, optics, detector, exp_time, n_frames, s
         and associated FITS header information.
     """
     sim_scene = optics.get_host_star_psf(scene)
-    if hasattr(scene, 'point_source_dra') or hasattr(scene, 'point_source_ddec'):
+    if hasattr(scene, 'point_source_dra') and len(scene.point_source_dra) > 0:
```
maxwellmb, line 42: I think we want a similar check to see if there's a host star, right?

maxwellmb (changes requested):
One small comment

corgisim/observation.py
zjwhale: Here, we only need to check whether companion properties have been provided for companion injection. A host star is not necessarily required, since we want to keep the capability to simulate a companion without a host star.

maxwellmb: Why would you call get_host_star_psf if you don't have a host star?

maxwellmb (changes requested):

corgisim/observation.py
zjwhale: We can not add a condition around `optics.get_host_star_psf(scene)` here, because the empty image with zero values is currently created inside `optics.get_host_star_psf()` itself.  If the inputt_scene has no host info, it will create a empty frame inside optics.get_host_star_psf() with:

```
if input_scene.if_empty:
            out_dim = self.grid_dim_out * self.oversampling_factor if self.return_oversample else self.grid_dim_out
            if self.cgi_mode == 'excam' and self.prism in ['POL0', 'POL45']:
                image = np.zeros((2, out_dim, out_dim), dtype=float)
            else:
                image = np.zeros((out_dim, out_dim), dtype=float)

            sim_scene.host_star_image = outputs.create_hdu(image, sim_info=sim_info)
            return sim_scene
```

maxwellmb (changes requested):
Let's remove the functionality to create an empty scene from this function and make it its own function. As part of that let's take the sim_info code and make it its own function that then gets used by the new empty scene function and the host star function and the point source function (only if needed in each case).

PR #221, opened 2026-05-01
fix roll angle convention

zjwhale:
## Purpose

The convention used by `corgisim` to map sky-plane coordinates to EXCAM coordinates and apply the roll angle was incorrect. This PR fixes that convention.

The correct convention is:

- In the sky plane, east is 90 degrees counterclockwise from north.
- In the EXCAM plane, the positive x-axis is 90 degrees clockwise from the positive y-axis.
- When `roll = 0`, north is aligned with positive y, and east is aligned with negative x.
- A positive roll angle means the EXCAM x-y frame rotates counterclockwise relative to the sky-plane coordinates.

The definition is described here: [add https://collaboration.ipac.caltech.edu/pages/viewpage.action?pageId=223743972&spaceKey=romancoronagraph&title=PA_APER%2Band%2BROLL%2Bangle%2Bconventions]
Screenshot 2026-04-30 at 5 02 40 PM

## Summary of Changes

**Major updates** to `instrument.py` and `test_roll_angle.py`.

### DIFF
--- a/test/test_roll_angle.py
+++ b/test/test_roll_angle.py
@@ -25,8 +25,8 @@ def test_roll_imaging():
     #Define companion properties
     #Add one companions, one in FOV
     mag_companion = [20]
-    companion_x_pos = [3*49.3]
-    companion_y_pos = [-3*49.3]
+    companion_x_pos = [-3*49.3]
+    companion_y_pos = [3*49.3]
     
     #### simulate using corgisim
     host_star_properties = {'Vmag': Vmag, 'spectral_type': sptype, 'magtype':'vegamag'}
@@ -81,7 +81,7 @@ def test_roll_spec():
     #Define companion properties
     #Add one companions, one in FOV
     mag_companion = [20]
-    companion_x_pos = [740]
+    companion_x_pos = [-1*740]
     companion_y_pos = [740]
     
     #### simulate using corgisim
@@ -128,11 +128,11 @@ def test_roll_spec():
     sep2_slit = np.sqrt(x2_slit**2+y2_slit**2)
     PA2_slit =  np.rad2deg(np.arctan2(y2_slit, x2_slit))
 
-    dPA = PA1-PA2
+    dPA = PA2-PA1
     assert sep1 == pytest.approx(sep2,abs=0.1), (f"Separation check failed: expected  {sep1:.3f} mas, got {sep2:.3f} mas).")
     assert dPA == pytest.approx(roll_angle,abs=0.1), (f"Roll-angle check failed: expected  {roll_angle:.3f} degree, got roll={dPA:.3f}degree).")
 
-    dPA_slit = PA1_slit-PA2_slit
+    dPA_slit = PA2_slit-PA1_slit
     assert sep1_slit == pytest.approx(sep2_slit,abs=0.1), (f"Separation for slit check failed: expected  {sep1_slit:.3f} mas, got {sep2_slit:.3f} mas).")
     assert dPA_slit == pytest.approx(roll_angle,abs=0.1), (f"Roll-angle for slit check failed: expected  {roll_angle:.3f} degree, got roll={dPA_slit:.3f}degree).")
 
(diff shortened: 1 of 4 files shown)


maxwellmb:
Although the full suite of tests didn't run, this did trigger "test_off_axis_source.py" to fail, and possibly test_roll_angle.py, but we're not sure if that second one failed because the test was cancelled (too long running tests).

test/test_roll_angle.py
```diff
@@ -109,24 +109,29 @@ def test_roll_spec():
     sim_scene_roll1 = optics_roll1.inject_point_sources(base_scene)
     x1, y1 = optics_roll1.optics_keywords_comp['source_x_offset_mas'],optics_roll1.optics_keywords_comp['source_y_offset_mas']
     sep1 = np.sqrt(x1**2+y1**2)
-    PA1 = np.rad2deg(np.arctan2(y1, x1))
-  
+    ###PA here is the angle relative to the +y axis, increasing towards +x axis.
+    PA1 = np.degrees(np.arctan2(x1, y1)) % 360
+    assert x1 == pytest.approx(-1*companion_x_pos[0],abs=0.1), (f"mapping check failed: expected  {-1*companion_x_pos[0]:.3f}, got roll={x1:.3f}).")
```
maxwellmb: I don't think the message here makes sense does it? why is it "got roll"?

maxwellmb (changes requested):

test/test_roll_angle.py
zjwhale: opps. I just fixed the typo.

maxwellmb:
I've lost track of the status of this and didn't quite follow the conversation today. 
@jrhom1, @semaphoreP, @zjwhale did I understand that we were going to move forward with the assumptions in this PR for now and will update if need be?

jrhom1:
@maxwellmb I think we're thinking of two different issues/PRs. I don't think we discussed this corgisim PR today but we did talk about the corgidrp issue related to this. Unfortunately, I think this PR is affected by the conversation that Alex and Neil are currently iterating on. I think from Neil's perspective, the roll angle convention should be the opposite sign of what is currently written into corgisim

zjwhale:
> @maxwellmb I think we're thinking of two different issues/PRs. I don't think we discussed this corgisim PR today but we did talk about the corgidrp issue related to this. Unfortunately, I think this PR is affected by the conversation that Alex and Neil are currently iterating on. I think from Neil's perspective, the roll angle convention should be the opposite sign of what is currently written into corgisim

Hi Jusin @jrhom1 is there change in the definition of roll angle convention since the discussion of in slack [here](https://romanspacetelescope.slack.com/archives/C065S7EDZ8X/p1777581868752969)? I think this PR comply with what we talk about in the thread. I also tag Neil @neilzim here for the discussion. 

If there is change, do you know is there any document for the new convention.

jrhom1:
@zjwhale the updates in the conversation have been in a direct message between me, Jason, Julia, Alex, and Neil. Alex and Neil are meeting tomorrow to reconcile the inconsistency. If Neil's interpretation is correct, I believe that means the convention is the opposite sign of what is defined on confluence pages and currently in corgisim.

zjwhale:
> @zjwhale the updates in the conversation have been in a direct message between me, Jason, Julia, Alex, and Neil. Alex and Neil are meeting tomorrow to reconcile the inconsistency. If Neil's interpretation is correct, I believe that means the convention is the opposite sign of what is defined on confluence pages and currently in corgisim.

Thanks for the heads up. Let me know once the new convention is finalized. I can update this PR accordingly.

jrhom1:
Hi @zjwhale , after discussing with Alex and Neil, we determined that PA_APER should be opposite in sign from the roll angle. Right now, it appears the default in corgisim is to assign the PA_APER keyword to be equal to input roll angle. Corgidrp has already been updated with this convention.

zjwhale:
> Hi @zjwhale , after discussing with Alex and Neil, we determined that PA_APER should be opposite in sign from the roll angle. Right now, it appears the default in corgisim is to assign the PA_APER keyword to be equal to input roll angle. Corgidrp has already been updated with this convention.

@jrhom1 Thanks for the updates. Is the PA_APER convention same as this plot? and roll angle is opposite to that?

Screenshot 2026-06-02 at 6 02 59 PM

jrhom1:
Sorry @zjwhale , I mixed up the conventions in my head again. I think the original convention is actually correct. PA_APER is the same sign as roll, and we fixed the DRP to handle this properly. Apologies for the confusion!

semaphoreP:
Note that corgisim term "ROLL" aligns with PA_APER. I forget if the IPAC definiton of "ROLL" also aligns with PA_APER (perhaps it is, as Justin alludes to above). But in general, we only define things in terms of PA_APER.

maxwellmb (approved):

PR #223, opened 2026-05-12
Negative satspots

### LINKED ISSUE #212: 1e-7 Satellite Spot Flux Inconsistency
In simulating 1e-7 commanded satellite spots for HLC-NFOV Band 1, the satellite spots themselves are not at consistent contrast values as commanded. For the 3e-8 DMs case (hlc_ni_3e-8), placing 1e-7 spots in the "+" configuration at 6.5 lambda/D (see below), this causes the DRP to fail at L3 to L4, as the "find_star" function is not able to converge on fitting for satellite spot positions. Specifically, the top spot is notably fainter than 1e-7 while some spots are notably brighter. The 5e-9 case successfully runs through the DRP to L4.

Images created by median combining non-spot images and subtracting them from the median-combined sat spot images.

**hlc_ni_3e-8**

Simulated Science Observation

Image

Simulated Reference Observation (note masking of central diffraction feature as a cosmic ray)

Image

**hlc_ni_5e-9**

Simulated Science Observation

Image

Simulated Reference Observation

Image

maxwellmb:
Add the option to include negative sat spots. A new test was added to test_add_satellite_spots.py, but is not necessary for test_minimal. 

- [x] I have verified that test_minimal.py passes and I have added a test there if appropriate.

## Checklist before merging

### DIFF
--- a/corgisim/instrument.py
+++ b/corgisim/instrument.py
@@ -1267,11 +1267,14 @@ class CorgiOptics():
         This function modifies the deformable mirror settings stored in `self.optics_keywords['dm1_v']` 
         by injecting satellite spots according to the provided `satspot_keywords`.
 
+        An optional "sign" keyword allows a user to define whether the satellite spots are added with a 
+        positive or negative cosine pattern on the DM. Defaults to "positive" if not specified.
+
         Parameters:
         ----------
         satspot_keywords : dict
             Dictionary specifying the parameters needed to define and inject 
-            satellite spots (sep_lamD, angle_deg, contrast, wavelength_m).
+            satellite spots (sep_lamD, angle_deg, contrast, wavelength_m, sign(options)).
 
         Returns:
         -------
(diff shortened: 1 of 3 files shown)


jrhom1:
#212

maxwellmb:
It's interesting to look at how things work for 1e-7 spots with the default DM pattern: 
image (32)

You can see that if you only have positive or negative, then you are subject to interference between the sat spots and the speckles, but when you use them both, then that goes away.

jrhom1:
I can also confirm that this settles issues the DRP had in finding all four sat spots in the 3e-8 raw contrast scenario. In fact, if you simulate a dataset that at least has the positive and negative sat spot frames, the DRP can find all four after combining the sat spot frames without any changes to DRP.

Although, there will probably need to be a change to the DRP related to this anyways since it currently uses the non sat spot science data to subtract the background from the sat spot frames. I think the FDD states that this background should be constructed from frames collected during the sat spot sequence that do not have any spots

maxwellmb:
Yep, I'm reviewing a DRP change from Jason to this effect.

zjwhale:
The PR looks good to me. The complete test is running now. It is ready to merger after the complete test passed.

zjwhale:
the complete test passed for this PR, I will approve the merge

zjwhale (approved):

PR #225, opened 2026-05-20
fix_noise_free_frame_saving_bug

zjwhale:
Fix the bug in saving the noise-free frame and add a test file


maxwellmb (changes requested):
Let's take this new test and merge it with test_output_1024.py into a new test_output.py.

zjwhale:
> Let's take this new test and merge it with test_output_1024.py into a new test_output.py.

I have combined the two test files.

maxwellmb (approved):

PR #226, opened 2026-05-20
Update requirements.txt

kjl0025:
Updating requirements.txt so that a version of emccd_detect is not specified (so that the latest emccd_detect will be pulled upon fresh installation).

- [x] I have verified that test_minimal.py passes and I have added a test there if appropriate.

## Checklist before merging
- [x] I have checked the complete tests pass on that branch (check the Github Actions tab)


kjl0025:
In addition to allowing for the latest version of emccd_detect, I also added in 4 spectral types (ones that are used by CAR-121 pupil images for k gain and nonlinearity calibrations), and I added in 2 keywords for emccd_detect, allowing a user to simulate nonlinearity and to simulate nonuniformity correctable by a master flat.

requirements.txt
```diff
@@ -4,6 +4,6 @@ scipy
 ipython
 pytest
 synphot==1.5.0
-emccd_detect==2.2.5
+emccd_detect
```
zjwhale: Do we need to set emccd_detect=2.2.6 or emccd_detect>2.2.5? My understanding is that emccd_detect does not have smearing effect in previous version and could cause errors if we use previous version.

zjwhale:
Hi  Kevin @kjl0025 , I leave a minor comment about the version of emccd_detect in requirement.txt. Other changes look good to me.

requirements.txt
kjl0025: True.  I'll put >=2.5.0.  2.5.0 is the earliest version with smear simulation.

helloWorlds0:
Since @kjl0025 took into account your comment @zjwhale , I'm approving this review and merging

helloWorlds0 (approved):

PR #230, opened 2026-05-28
fix bugs

zjwhale:

## Checklist before merging


helloWorlds0:
This fixes the bugs introduced by https://github.com/roman-corgi/corgisim/pull/217 right? Since we were using fpm and nd filters combination that are no longer allowed, the tests fail.  I will merge it directly so that the failed tests don't hide genuine problems.

helloWorlds0 (approved):

Issue #231, opened 2026-05-29
Setting a new VISITID in save_hdu_to_fits does not propagate through to filename


juliamilton:
### Description

`save_hdu_to_fits` has an `overwrite_pri_keywords` argument which is used to override primary header values before saving. However, when `write_as_L1=True`, the filename is derived from `prihdr['FILENAME']` that was set before the overrides are applied. This means any keyword that setting a different VISITID is not propagated to the saved filename even if correctly passed in via `overwrite_pri_keywords`.

### Steps to Reproduce

outputs.save_hdu_to_fits(
    hdul,
    outdir=outdir,
    write_as_L1=True,
    overwrite=True,
    overwrite_pri_keywords={'VISITID': '0200001001001001003'}
    )

The saved file will still use the original VISITID, not the overridden value.

zjwhale:
Thanks for catching this. I will work on it.

PR #240, opened 2026-06-13
add feature to input visit_ID

zjwhale:
## Describe your changes and reference any relevant issues #231

This PR checks and fixes the issues related to overwriting `VISITID` in the headers, as described in issue #231. A new test has been added to verify that `VISITID` can be overwritten correctly. In addition, the corresponding header keywords, `PROGNUM`, `EXECNUM`, `CAMPAIGN`, `SEGMENT`, `OBSNUM`, and `VISNUM`, are now automatically updated when `VISITID` is overwritten.

This PR also adds a new input keyword, `visit_id`, to `Corgi.Optics`, which is passed into the headers. This allows users to provide `visit_id` from the beginning, instead of overwriting it later.

## Checklist before merging


maxwellmb:
Note: Jingwen and I agree that this doesn't actually resolve #231 (but it's still a useful feature)

corgisim/outputs.py
```diff
@@ -190,6 +210,36 @@ def save_hdu_to_fits( hdul, outdir=None, overwrite=False, write_as_L1=False, fil
         overwrite_pri_keys = overwrite_pri_keywords.keys() if overwrite_pri_keywords else []
         overwrite_ext_keys = overwrite_ext_keywords.keys() if overwrite_ext_keywords else []
 
+        if "VISITID" in overwrite_pri_keys:
```
maxwellmb: Also need to update the filename here

corgisim/outputs.py
```diff
@@ -190,6 +210,36 @@ def save_hdu_to_fits( hdul, outdir=None, overwrite=False, write_as_L1=False, fil
...
+
+            if not re.fullmatch(r"\d{19}", visit_id):
+                raise ValueError(
+                    "VISITID must be a 19-character string in the format "
+                    "'PPPPPEEECCCNNNNVVV' "
+                    "(P=Program, E=Execution, C=Campaign, N=Observation, V=Visit)."
+                )
+
+            # If VISITID is overwritten, the corresponding keywords must also be updated.
+            prognum  = visit_id[0:4]
```
maxwellmb: just put these directly into the keywords below

corgisim/outputs.py
```diff
@@ -190,6 +210,36 @@ def save_hdu_to_fits( hdul, outdir=None, overwrite=False, write_as_L1=False, fil
...
+            execnum  = visit_id[4:7]
+            campaign = visit_id[7:10]
+            segment  = visit_id[10:13]
+            obsnum   = visit_id[13:16]
+            visnum   = visit_id[16:19]
+
+            overwrite_pri_keywords.update({
+                "PROGNUM": prognum,
```
maxwellmb: let's add a check that forbids overwriting PROGNUM, EXECNUM, etc. Instead it _has_ to be VISITID that is modified and those changes propagate.

maxwellmb (changes requested):

zjwhale:
I have addressed all he comments. This PR now fix the bugs from issue#231 and add the option to input visit ID to corgi.optics. In addition, I added new condition that overwrite keyword function in `save_hdu_to_fits()` can only applied to L1 products . Because intermediate products only have comment headers.

helloWorlds0:
This look good, except i don't think visit_id should be in the Optics class (nor, come to think of it, vistype). It's not a property of the optics.

zjwhale:
> This look good, except i don't think visit_id should be in the Optics class (nor, come to think of it, vistype). It's not a property of the optics.

Yeah, I see your point. But I don't think we have better place to put. In ideal, it could be put in `observation.py.` But we don't have a way to pass  values into headers there. Do you have any suggestion?

helloWorlds0:
> > This look good, except i don't think visit_id should be in the Optics class (nor, come to think of it, vistype). It's not a property of the optics.
> 
> Yeah, I see your point. But I don't think we have better place to put. In ideal, it could be put in `observation.py.` But we don't have a way to pass values into headers there. Do you have any suggestion?

I see two options: 
Option 1: We have a "Sequence" or "Campaign" class that would have this sort of features. Conceptually it would probably be the cleanest way to do it but it's a bit of work to conceptualize and implement
Option 2: Can't we just keep the part where the visit_id value is reported to the subsequent headers, but the only way to change it is through the overwrite_headers functions?

zjwhale:
> > > This look good, except i don't think visit_id should be in the Optics class (nor, come to think of it, vistype). It's not a property of the optics.
> > 
> > 
> > Yeah, I see your point. But I don't think we have better place to put. In ideal, it could be put in `observation.py.` But we don't have a way to pass values into headers there. Do you have any suggestion?
> 
> I see two options: Option 1: We have a "Sequence" or "Campaign" class that would have this sort of features. Conceptually it would probably be the cleanest way to do it but it's a bit of work to conceptualize and implement Option 2: Can't we just keep the part where the visit_id value is reported to the subsequent headers, but the only way to change it is through the overwrite_headers functions?

Sorry for taking a while to get back to this PR. I think we do want to have `visit_id` as an input from the beginning, rather than only overwriting it in `output.save_hdu_to_fits()`, for two reasons:

1. It's cleaner for users to specify a different `visit_id` for each visit.
2. We have other headers and filenames depend on `visit_id` , so updating it only at the end could lead to inconsistencies.

The proposed `Sequence` or `Campaign` class could be a nice addition in the future, but I don't think it's a high priority right now.

Overall, I think the current implementation is good to accept as it is, and we can revisit the `Sequence`/`Campaign` option in the future if needed.

helloWorlds0:
> It's cleaner for users to specify a different `visit_id` for each visit.

I don't see how? If the user case if that they are using, for instance, generate_observation_sequence, then we can add the option to specify the visit_id and vis_type there (if that the use case?)

>We have other headers and filenames depend on visit_id , so updating it only at the end could lead to inconsistencies.

I see that PROGNUM, EXECNUM, SEGMENT, OBSNUM, VISNUM, and FILENAME depend on visit_id, and when visit_id is overwritten, they are updated. Are there others that I have missed?

maxwellmb (approved):

PR #241, opened 2026-06-13
fix bugs in main branch

zjwhale:

## Checklist before merging

### DIFF
--- a/corgisim/outputs.py
+++ b/corgisim/outputs.py
@@ -195,23 +195,29 @@ def save_hdu_to_fits( hdul, outdir=None, overwrite=False, write_as_L1=False, fil
                 hdul[0].header[key] = overwrite_pri_keywords[key]
             else:
                 raise KeyError(f"Primary header keyword '{key}' not found in HDUList.")
-        
-        for key in overwrite_ext_keys:
-            if key in hdul[1].header:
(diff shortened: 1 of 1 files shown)


corgisim/outputs.py
```diff
@@ -195,23 +195,30 @@ def save_hdu_to_fits( hdul, outdir=None, overwrite=False, write_as_L1=False, fil
...
-            if key in hdul[1].header:
-                hdul[1].header[key] = overwrite_ext_keywords[key]
+
+        if overwrite_ext_keywords is not None:
+            if write_as_L1:
+                for key in overwrite_ext_keys:
+                    if key in hdul[1].header:
+                        hdul[1].header[key] = overwrite_ext_keywords[key]
+                    else:
+                        raise KeyError(f"Extension header keyword '{key}' not found in HDUList.")
```
maxwellmb, line 205: why does this one need the `if write_as_L1`?

zjwhale: if `write_as_L1` is false, it is to save intermediate products, not L1 fits files. In that case, the hdu only have a primary header, no extension header. So `hdul[1]` will cause error. This is to give an error message if an user want to overwrite the extension header for non-L1 products.

maxwellmb (approved):

PR #243, opened 2026-06-19
add RA and DEC to scene

zjwhale:
## Describe your changes and reference any relevant issues 242

This PR add input options to scence for target RA and DEC. These values will be populated headers. One thing to check is that current confluence page says the unit for RA ans DEC is in mas, whcih is not commonly used for them. 

Updates: check with Alex that we need to use degree for RA and DEC. This has been implemented in this PR

## Checklist before merging


maxwellmb:
@agreenbaum, are the planned units for RA and DEC degrees? Or will they be a strings (Sexagesimal for RA)?

zjwhale:
> @agreenbaum, are the planned units for RA and DEC degrees? Or will they be a strings (Sexagesimal for RA)?

Yes, I have confirmed with Alex that it should be in degrees

maxwellmb (approved):

PR #246, opened 2026-06-25
Add time series from CPGS based on os11

AlexBidot:

- New wavefront_estimation module to retrieve the wavefront error from OS11 scenario
- Added to the existing observation.py module the possibility to include speckle time variation in the main function generate_observation_scenario_from_cpgs. The limitation is that the total exposure time shall not be longer to the one from os11. Added some
- New Jupyter tutorial to generate a time series in 2 different manners (either from scratch or from CPGS xml file).
if the time series are generated "from scratch", the generated sequence will be divided the same as os 11 with alternating the roll angle and observation cycle between reference and target star. Outputs will be saved both with and without the emccd noise.

## Checklist before merging


corgisim/observation.py
```diff
@@ -2,10 +2,90 @@
 ## This will likely contain functions simmilar to the functionality in Jorge's corgisims_obs 
 import corgisim
 import os
-from corgisim import scene, instrument, inputs, observation, outputs
-import copy 
+from corgisim import scene, instrument, inputs, observation, outputs, wavefront_estimation
+import copy
 
-def generate_observation_sequence(scene, optics, detector, exp_time, n_frames, save_as_fits= False, output_dir=None, full_frame= False, loc_x=None, loc_y=None):
+def check_cpgs_os11_compatibility(visit_list):
```
helloWorlds0, line 8: If I understand corrrectly, the compatibility needs in this function are :

- 	only one angle for the reference star (this might be problematic if the user chooses the Constant Roll Angle option, because we take the position angle from the cpgs file, and a constant roll angle means that the position angle is likely to vary slightly between visits) 

- 	only 4 target visits between each reference star visits.

But there are extra condition from wavefront estimation: 

- 	no more than 6 reference star visit

- 	each visit must have a total time inferior to the corresponding visit in OS11 (around 2 hours I think?)

If we have a function to test compatibility, shouldn't we check all compatibility criteria here?

examples/time_series_demo.ipynb
helloWorlds0, line 1: Would be worth it to have a demo that's a bit more illustrative (compatibility info, figure for OS 11 scenario, images with/without the drift...). There are also extra steps to have the files in example show up in the doc but we can do that later

corgisim/observation.py
```diff
@@ -31,13 +111,18 @@ def generate_observation_sequence(scene, optics, detector, exp_time, n_frames, s
...
+        zernike_value_m (numpy 2d array (n_index, n_frames), optional): List of the zernike wavefront aberration values in meters.
+            If None, no speckles variation will be generated.
 
     Returns:
         list[corgisim.scene.SimulatedImage]: A list of :py:class:`corgisim.scene.SimulatedImage` objects,
         where each object represents a single generated observation frame with its image data
         and associated FITS header information.
     """
+
     sim_scene = optics.get_host_star_psf(scene)
```
helloWorlds0, line 126: Considering that including drift will make simulating much more time consuming (which is normal), it might not be a big impact, but still: When we include the zernike, we generate the sim scene once more than we need. We can add a condition here to prevent that.

corgisim/wavefront_estimation.py
```diff
@@ -0,0 +1,326 @@
...
+
+    sep_mas = np.sqrt(x_off_mas**2 + y_off_mas**2)
+    para_ang = np.arctan2(y_off_mas, x_off_mas)
+
+    x_off_mas_roll = sep_mas * np.cos(para_ang + theta_rad)
+    y_off_mas_roll = sep_mas * np.sin(para_ang + theta_rad)
+
+    return x_off_mas_roll, y_off_mas_roll
+
+def get_optimal_emgain(noiseless_image_hdu):
```
helloWorlds0, line 162: good rapid-check to have, there is an issue on corgisim about using eetc to get the optimal exposure/gain from eetc, maybe let's add a TODO here

corgisim/wavefront_estimation.py
```diff
@@ -0,0 +1,326 @@
...
+                    output_save_file = f'psf_target_{n}_cycle_{cycle}_roll_{roll}.fits'
+                    output_ccd_save_file = f'psf_target_{n}_cycle_{cycle}_roll_{roll}_ccd.fits'
+
+                    if point_source_info is not None:
+                        # compute the x_off, y_off of the companion with respect to roll angles
+                        point_source_info_roll = point_source_info
+
+                        if roll % 2 == 0:
+                            for i in range(len(point_source_info)):
+                                point_source_info_roll[i]['position_x'], point_source_info_roll[i]['position_y'] = estimate_companion_position_roll(point_source_info[i]['position_x'], point_source_info[i]['position_y'],
```
helloWorlds0, line 284: Not sure why this is used instead of having the roll in the optics instrument (I am guessing because it was written before the roll angle got included?). Anyway we probably should just change the roll in the optic

corgisim/wavefront_estimation.py
```diff
@@ -0,0 +1,326 @@
...
+        t0, t1 = get_reference_cycle_times(cycle)
+    else:
+        raise ValueError("obs must be either 'science' or 'ref'")
+
+    # check that the requested number of exposures is compatible with OS11 data
+    t_requested = exp_time * n_frames  # sec - requested time
+    if t_requested > (t1 - t0):
+        raise ValueError('Number of frames and individual exposure times are not compatible with the OS11 scenario')
+
+    noll_index = np.arange(4, 46, 1)
```
helloWorlds0, line 114: It seems to me that the zernike index are a constant, so we don't need to calculate them each time

helloWorlds0 (changes requested):
There is no test. Generally, it's worth asking 'How do I know my code works? What would it look like if it didn't work?" as a way to get started with tests

corgisim/observation.py
maxwellmb: +1, yes, I agree that we should add Sophie's additional suggested tests.

corgisim/observation.py
```diff
@@ -46,6 +131,12 @@ def generate_observation_sequence(scene, optics, detector, exp_time, n_frames, s
     
     if full_frame == False :
         for i in range(0, n_frames):
+            if zernike_index is not None:
+                optics.optics_keywords.update({'zindex': zernike_index, 'zval_m': zernike_value_m[i]})
```
maxwellmb, line 135: If this is how zernike_value_m gets indexed, then should the docstring say "(numpy 2d array (n_frames, n_index)" instead of "(numpy 2d array (n_index, n_frames)"?

corgisim/observation.py
maxwellmb: +1

corgisim/wavefront_estimation.py
```diff
@@ -0,0 +1,326 @@
...
+        time_ref_cycle_min, time_ref_cycle_max (floats): start/end times of the requested reference star cycle
+
+    """
+
+    if (cycle < 1) or (cycle > 6):
+        raise ValueError('Cycle number for the reference star in the OS11 scenario must be an integer between 1 and 6')
+
+    data_dir = os.path.dirname(corgisim.lib_dir)
+
+    batch_info = pd.read_csv(os.path.join(data_dir , 'test/test_data/OS11inputs.csv'))
```
maxwellmb, line 31: If this is now being used in the main code, I think we should move it out of test/test_data

corgisim/wavefront_estimation.py
```diff
@@ -0,0 +1,326 @@
...
+import matplotlib.pyplot as plt
+import numpy as np
+import proper
+import roman_preflight_proper
+from scipy.interpolate import interp1d
+from corgisim import outputs
+import os
+
+
+def get_reference_cycle_times(cycle: int = 1):
```
maxwellmb, line 14: overall we haven't been defining the input types this way. In order to keep the style the same across the package, please remove it. (Perhaps one day we'll migrate to the new formatting...)

corgisim/observation.py
maxwellmb: I also wonder if we want to have options to loosen the restrictions a bit. For example, if the reference were at a different roll, or if we _did_ want more than 6 reference visits per star. Perhaps let's make an issue about that as a future feature. Yes...it comes at the cost of fidelity..but maybe sometime we want that.

corgisim/wavefront_estimation.py
```diff
@@ -0,0 +1,326 @@
...
+        t0, t1 = get_science_cycle_times(cycle, roll)
+    elif obs == 'ref':
+        t0, t1 = get_reference_cycle_times(cycle)
+    else:
+        raise ValueError("obs must be either 'science' or 'ref'")
+
+    # check that the requested number of exposures is compatible with OS11 data
+    t_requested = exp_time * n_frames  # sec - requested time
+    if t_requested > (t1 - t0):
+        raise ValueError('Number of frames and individual exposure times are not compatible with the OS11 scenario')
```
maxwellmb, line 112: Perhaps you can add details to the error message. How long was requested, and what is the maximum for OS11? (But as Sophie suggested, lets move this to the compatibility check at the start.

corgisim/observation.py
maxwellmb: indeed from wavefront_estimation.py it looks like its (n_frames,n_index)

corgisim/wavefront_estimation.py
```diff
@@ -0,0 +1,326 @@
...
+            a_lowfs = f_lowfs_corr_interp(t_interp)
+            a_interp += a_lowfs
+
+        for k in range(n_frames):
+            zval_m[k, i] = a_interp[k]
+
+    return noll_index, zval_m
+
+
+def estimate_companion_position_roll(x_off_mas: float, y_off_mas: float, roll_0: str = 'positive', theta: float = -13):
```
maxwellmb, line 136: I'm not sure why this is in wavefront_estimation.py

corgisim/wavefront_estimation.py
```diff
@@ -0,0 +1,326 @@
...
+    """
```
maxwellmb, line 137: Please add a description of this function to the docstring

corgisim/wavefront_estimation.py
```diff
@@ -0,0 +1,326 @@
...
+    sep_mas = np.sqrt(x_off_mas**2 + y_off_mas**2)
+    para_ang = np.arctan2(y_off_mas, x_off_mas)
+
+    x_off_mas_roll = sep_mas * np.cos(para_ang + theta_rad)
+    y_off_mas_roll = sep_mas * np.sin(para_ang + theta_rad)
+
+    return x_off_mas_roll, y_off_mas_roll
+
+def get_optimal_emgain(noiseless_image_hdu):
+    """
```
maxwellmb, line 163: please add a description of this function to the docstring

corgisim/wavefront_estimation.py
```diff
@@ -0,0 +1,326 @@
...
+    noiseless_image = noiseless_image_hdu.data
+    saturation = 12000
+    threshold = 0.8 * saturation
+    max = np.nanmax(noiseless_image)
+    emgain = threshold/max
+
+    return emgain
+
+def generate_time_series_from_os11(ref_star_properties, target_star_properties, point_source_info, frame_exp, total_exp_time, outdir=None, cor_type='hlc', bandpass = '1F', polaxis=10, dm_case='3e-8'):
+    """
```
maxwellmb, line 181: This function looks like it's only really used as an example in the notebook, and only really generates one specific scenario. You could either a) put it in its own .py in examples/ or just take the innards of the function and have them be explicitly part of the tutorial. Either way I think it needs to be moved from here along with the two options above, since it's more of a use-case rather than core wavefront estimation code.

maxwellmb: If you do keep it, please add a description to the docstring.

examples/time_series_demo.ipynb
maxwellmb: I agree, it would be useful to have a few more details in the tutorial, with a bit more of a narrative description of what the time-variable feature does, its use-cases and its limitations. Have some illustrative figures could help too. 

Also, please have the notebook fully run with outputs, before committing.

maxwellmb (changes requested):
Thanks for your work on this! I added a few comments. Overall structurally i think we're mostly there. As @helloWorlds0 pointed out though, we require some basic functionality tests to show that your new features are doing what we expect them to do.

PR #259, opened 2026-07-29
207 update cpgs parser

### LINKED ISSUE #207: Update CPGS parser
The new CPGS version is online, so I'll be updating the CPGS parser

helloWorlds0:
Main change is that we can now handle the new CPGS files. 
We have satellite spots and polarimetry mode, different bands, NFOV,  WFOV, etc. We also now create one folder per visit (see image) 
What we specifically don't have: 
- an integrated call to eetc in case of autogain
-  scaling the speckle field to the contrast that we want 
- any test on the spec mode. I started on the parsing of the spec mode, but I haven't put any spec cpgs file in the test. Basically this doesn't have spectroscopy
- A tutorial 

It's also farily light on tests: I have tested that it runs, but not that it creates the image that we want. I plan to add tests later, but I think this can start to be reviewed. 
image

## Checklist before merging


corgisim/observation.py
```diff
@@ -88,23 +99,116 @@ def generate_observation_scenario_from_cpgs(filepath, save_as_fits= False, outpu
...
         host_star_properties = {'Vmag': scene_target._host_star_Vmag, 'spectral_type': scene_target._host_star_sptype, 'magtype': scene_target._host_star_magtype, 'ref_flag': False}
         scene_target = scene.Scene(host_star_properties, point_source_info)
 
+    #Satellit spot configuration
+    satspots_are_present = (satellite_dict_target is not None) 
+    if satellite_dict_target is not None:
+        contrast1 = 1e-7
+        contrast2 = 1e-5
+        sep1 = 6.25
+        sep2 = 13
```
zjwhale: These four parameters are hard-coded here. Are they always expected to have these values? Or should they be passed in from the CPGS inputs/ other input configuration instead?

zjwhale: It would be more flexible to make these configurable attributes, with these numbers as the default values, in case we need to use different satellite-spot separations or contrasts in some scenarios.

zjwhale (commented):
TThe overall PR looks good to me. I left a few minor comments. 

As Sophie and I discussed during the in-person meeting, we agreed to merge this PR as-is and make the following change in a separate PR:

We should add an option to generate satellite-spot frames directly from generate_observation_sequence(). The idea is to add an additional boolean keyword indicating whether the sequence being generated is for satellite-spot frames. If true, the function would generate N * 3 frames, where N is the desired number of satellite-spot frames. Then generate_observation_scenario_from_cpgs() can call generate_observation_sequence() for the satellite-spot frames.

I will create an issue to track this.

corgisim/observation.py
helloWorlds0: CPGS gives us the name of a configuration, not these parameters. They come from this part of the CPGS form 
image

helloWorlds0: So if, for instance, TDD_SATELLITE_SPOTS_SLOT0 changes definition, we'll need to update it, but there really isn't a way around it

zjwhale: I see. I will start a full test run and if it is pass, I will merge this PR

zjwhale: Just saw you already did it.

zjwhale (approved):
looks good

PR #261, opened 2026-07-31
Update default bias

helloWorlds0:

## Checklist before merging

### DIFF
--- a/corgisim/instrument.py
+++ b/corgisim/instrument.py
@@ -1640,7 +1640,7 @@ class CorgiDetector():
             - full_well_serial (float, optional): full well for serial register; 90K is requirement, 100K is CBE
             - dark_rate (float, optional): Dark current rate, e-/pix/s; 1.0 is requirement, 0.00042/0.00056 is CBE for 0/5 years
             - cic_noise (float, optional): Clock-induced charge noise, e-/pix/frame; Defaults to 0.01.
-            - read_noise (float, optional): Read noise, e-/pix/frame; 125 is requirement, 100 is CBE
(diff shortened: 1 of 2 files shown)


corgisim/instrument.py
```diff
@@ -1641,7 +1641,7 @@ def define_EMCCD(self, emccd_keywords=None ):
             - dark_rate (float, optional): Dark current rate, e-/pix/s; 1.0 is requirement, 0.00042/0.00056 is CBE for 0/5 years
             - cic_noise (float, optional): Clock-induced charge noise, e-/pix/frame; Defaults to 0.01.
             - read_noise (float, optional): Read noise, e-/pix/frame; 125 is requirement, 100 is CBE
-            - bias (int, optional): Bias level (in digital numbers). Defaults to 0.
+            - bias (int, optional): Bias level (in digital numbers). Defaults to 1500.
```
zjwhale, line 1644: Thanks for the updates. Could you point to the source for the default bias value of 1500? I'm not suggesting it's incorrect; I'd just like to know where that value comes from.

zjwhale: @kjl0025 Hi Kevin, could you confirm the number and is that only for pc mode? Thanks!

kjl0025: 1500 is a good number for bias in general, for both PC and analog.  Also, I notice that line 1643 above says that 100 is CBE.  That should be 165 for the CBE.

helloWorlds0: it is 165 in the code, changed the comment to reflect that

kjl0025 (approved):

Issue #263, opened 2026-08-06
Support polarized on-axis host stars


toshiyukimizuki:
Currently, `pol_state` is applied only to companion sources (inject_point_sources). There is no option to define polarization for the host star.
Please add a separate host-star polarization keyword and apply it in `get_host_star_psf()`. The default behavior should remain unpolarized.

PR #265, opened 2026-08-12
Simulate polarized host stars

everlastingEric:
Address issue #263, added the ability to provide a Stokes vector to the host star and simulate it in polarized light.
- Added the function `obtain_polarized_intensities_from_fields`  in `CorgiOptics` which now handles all polarimetric simulations of point sources, including both on-axis stars and off-axis companions, and on-axis stars with finite diameter or jitter.
- Updated the polarimetry demo notebook to include the new polarized host star parameters.
- Updated `test_polarimetry.py` to also simulate a polarized host star for testing.

## Checklist before merging


corgisim/instrument.py
```diff
@@ -520,33 +525,22 @@ def calculate_images_tem_for_pol0_or_pol45(N_offsets_counting_origin,delta_e_lib
...
-            
-        elif self.prism == 'POL45':
-            #45/135 case
-            # The four electric field components are:
-            # delta_E_m45in_45out
-            # delta_E_45in_45out
-            # delta_E_m45in_135out
-            # delta_E_45in_135out
-            # Set up a library of keys to identify these components:
-            delta_e_keys = {'1':'delta_E_m45in_45out','2':'delta_E_45in_45out','3':'delta_E_m45in_135out','4':'delta_E_45in_135out'}
```
maxwellmb: I think we still need these, no?

corgisim/scene.py
```diff
@@ -40,6 +40,8 @@ class Scene():
             - "ref_flag" (boolean):optional, whether the input scene is a reference star (True) or a science target (False). Default is false
             - "stellar_diam_mas" (float): The stellar diameter of the host star in mas.
             - "target_name" (str): optional, the name of the target star, used for headers. Default is UNKNOWN.
+            - "pol_state" (float array): optional, vector of length 4 consisting of the I, Q, U and V components of the stokes parameter
```
maxwellmb: why don't we just call it "stokes_vector" so that it's unambiguous as to what it is.

corgisim/scene.py
```diff
@@ -40,6 +40,8 @@ class Scene():
...
+                describing how the source light is polarized, default is unpolarized or [1,0,0,0]
```
maxwellmb: We should be clear: do we expect a normalized stokes vector? Such that I is always 1 and the next three components are fractional?

maxwellmb (changes requested):
A couple comments in the code. 

Also, I think we probably want a test of the speckle field (w/o a companion) having the expected polarization signature (averaged across the focal plane). It seems like right now only a companion is tested.

corgisim/instrument.py
everlastingEric: I think with the way I updated this, it doesn't matter if the focal plane electric field is already aligned with the Wollaston directions, since everything gets converted to a Mueller matrix image in order to simultaneously model partially polarized sources while keeping the polarization-dependent speckle field distributions intact. So as a result, I've kept the computation exactly the same for both prism configurations, and the only difference is at the very end where the polarizer Mueller matrices that gets applied are different.

corgisim/scene.py
everlastingEric: I think so yes, I will make it clear in the docstring

